PR #67, opened 2019-07-14
feat(indentation): allow spaces after tab

### LINKED ISSUE #66: Support "tabs for indentation, spaces for alignment" style
This is a continuation of discussion of the [issue in editorconfig-checker.javascript project](https://github.com/editorconfig-checker/editorconfig-checker.javascript/issues/68).

"Tabs for indentation, spaces for alignment" is a common coding style. It allows the code viewers to adjust any tab space they want without breaking alignments. Addition information is presented in these articles: https://dmitryfrank.com/articles/indent_with_tabs_align_with_spaces, https://vim.fandom.com/wiki/Indent_with_tabs,_align_with_spaces. This style is [called "Smart tabs"](https://www.jetbrains.com/help/phpstorm/settings-code-style-tabs-and-indents-common.html#f2a9d78e) in some IDEs (e. g. in JetBrains IDEs).

For now for projects using this style, `editorconfig-checker` shows the following error for each line with alignment:

```
136: Wrong indentation type(spaces instead of tabs)
```

I see several ways to deal with it it.

I think, the most natural way is automatically detect the alignment with spaces. If we are in a block, nested with N tabs, and then if we see spaces after N tabs, we should consider it an alignment with spaces.

Another approach would be to introduce additional command line key to enable this style or additional custom parameter to `.editorconfig` (if this is possible).

mstruebing:
TODO:
- [x] adjust tests to test correctly for both variants (with spaces allowed and without)
- [x] use options/params parameter for better maintainability instead of simply passing down more props

closes #66


okalachev:
Are you sure it's going to be OK, if we would not consider the blocks nesting, and just consider any spaces after tabs an alignment?

mstruebing:
Yes, I mean otherwise I would need to implement a nesting/indentation logic for _EVERY_ programming language and that isn't really handy to do. Also, for such cases, there are linters for every programming language. This is only a linter to check if files are respecting the editorconfig rules.

Or what do you think about that?

mstruebing:
hey @okalachev,
I've implemented a solution, do you think this would work as you've expected?

okalachev:
@okalachev, I've built this version. Spaces indentation detection works great!

I've found two problems though.

With this modification this kind of indentation errors are not detected:

Screenshot 2019-07-15 at 15 00 32

Also, first level alignments are still considered errors:

Screenshot 2019-07-15 at 15 06 30

Unfortunately I don't see an obvious way to fix them.

mstruebing:
The first one should be correct because you have a tab first and use spaces to indent it. 
The second one shouldn't be allowed, because it starts with spaces and not with a tab.
That's how I've understood the desired behaviour.

If you want to have the behaviour of your screenshots provided by `editorconfig-checker` I sadly must say that I will not implement this.
This means I need to program a full language parser(for every programming language btw) to check the level of needed indentation. As editorconfig is generally language agnostic and shouldn't care which language or markup or whatever text is provided this would mean it should do things it is not supposed to do.

"So much complexity in software comes from trying to make one thing do two things." - Ryan Singer

okalachev:
I mean, the first screenshot is now false negative (after the change). Is it OK?

The second screenshot is false positive. And yes, there is no simple way to fix it (only adding `editorconfig-disable-line`).

okalachev:
@mstruebing, I don't want this issue abandoned, please, answer, what do you think :)

mstruebing:
I could release a version in the current state which enables the behaviour for to alles spaces after tabs with a flag. Would this help you?

okalachev:
You mean to introduce a new command line flag, because of the false negative check?

mstruebing:
This one will be able to do with a flag: 
Screenshot 2019-07-15 at 15 00 32

This one will never be supported by editorconfig-checker: 
Screenshot 2019-07-15 at 15 06 30

okalachev:
What flag do you mean? Why we need a flag?

mstruebing:
Because I don't want the behaviour that spaces are allowed after tabs for every user as this is a special case and should not be the default behaviour.

The usage would be somehow like: `ec --allow-spaces-after-tabs`.

okalachev:
Ok, let’s make it like this, it would be helpful.

mstruebing:
Hey @okalachev - I've pushed the version to use with the flag now.
The flag to use is: `spaces-after-tabs`.
Could you give it a try if this is working as you would expect?
If so I would polish it a bit (see my first edited comment) and release a new version.

okalachev:
I tested it in my project, and it seems to work!

The only problem left (not related to this) is detecting "Wrong amount of left-padding spaces" in Python files, where spaces are used for aligning _and_ indentation. Is there any way to disable this individual check?

mstruebing:
unfortunately not right now, but I've thought about implementing this in the future.

okalachev:
As it's not implemented so far, I created a temporal fork with this check disabled.

I'm including this tool to my project: https://github.com/CopterExpress/clever/pull/149.

mstruebing:
feel free to add a proper PR if you feel comfortable doing it.

This PR here is ready this weekend I think. I plan to work on it today and maybe tomorrow.

mstruebing:
It's got finished, thanks for your input, time and patience :)

perlun:
For anyone stumbling into this PR in recent years :slightly_smiling_face:, the way to enable this setting is now the following flag in `.editorconfig-checker.json`:

```json
{
  "SpacesAfterTabs": true
}
```

(If you don't have a config file already, use `editorconfig-checker -init` to generate a default config which you can edit to suit your needs)

PR #92, opened 2019-10-25
Nix Package

vbrandl:
This PR enables building the checker as a nix package.

Three new targets were added to the Makefile:

* `make nix-build`: builds the nix package
* `make nix-install`: installs the nix package
* `make nix-update-dependencies`: updates the dependencies stored in `deps.nix`

Maybe it is possible to push the package into a [cachix](https://cachix.org/) cache, so others can install the package without having to build it themselves. There is a Github Action to to this: https://github.com/cachix/cachix-action

The version is now stored in the file `VERSION` in the root of the repository. At compile time, the value for `version` in `main.go` is set. This version is also used in the nix expression.


mstruebing:
Could you give a short description on how to use this? 
When running either of the new Makefile targets I get an error which says:

`error: file 'nixpkgs' was not found in the Nix search path (add it using $NIX_PATH or -I), at (string):1:13`

Arch Linux with freshly installed `nix-bin`  and not further configuration.

```
$ nix --version
nix (Nix) 2.3
```

Otherwise, great that you force me to think about nix, it's on my list since long but I didn't do anything about it.^^

vbrandl:
First of all, I would recommend installing nix as described here: https://nixos.org/nix/manual/#sect-multi-user-installation

This way nix manages (and updates) itself and you do not depend on the AUR package.

If you are using the AUR package, you need to add the `nixpkgs` channel first:

```
nix-channel --add https://nixos.org/channels/nixpkgs-unstable
nix-channel --update
```

After that, the commands in the Makefile should work.

mstruebing:
It is working, but it is dropping the binary in: `nix/store/j1r1w80gk13986q34fd6sx2yax46nh8m-editorconfig-checker-2.0.3-bin/bin/editorconfig-checker` 

Is there an easy way to add installed binaries to your path or symlink it to a position  in your path or something else?

vbrandl:
I guess it worked after you added the `nixpkgs` channel?

mstruebing:
Yep, I first tried the mentioned solution but this didn't work either because some symlinks were broken and stuff, so I decided to go again with `nix-bin` and `archlinux-nix` I additionally needed to run: `sudo chown -R $USER. /nix/var/nix/{gcroots,profiles}`

PR #107, opened 2020-05-04
Better makefile for distros

stephanlachnit:
I'm currently interested to pack editorconfig-checker for Debian, however there is no installation process in place. So I did this in the first commit by adding three targets: all (default build target), install and uninstall. I also used standard path names, using the default values even though they will be overwritten by most distros (e. g. `/usr` as prefix instead of `/usr/local`).
In the second commit I marked basically all targets except `bin/ec` as phony. Normally, make expects that the name of a target is the output file, but that's not the case here.
In the third commit I added a basic man page, mostly auto generated with a bit of optical lifting. You can view it with `man -l docs/editorconfig-checker.1`, note here that the `-config` option has a pretty bad description (taken from `bin/ec --help`), you may want to change it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 * Change `Dockerfile` `CMD` to allow the usage of `ec` everywhere inside the image [#105](https://github.com/editorconfig-checker/editorconfig-checker/pull/105) ([@chambo-e](https://github.com/chambo-e))
 ### Security
 ### Misc
+* Marked non-build targets in the Makefile as phony
 
 ## [2.0.3] - 2019-09-13
 ### Fixed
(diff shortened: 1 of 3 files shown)


stephanlachnit:
> What do you mean exactly by packaging it for debian? To provide a PPA?

I want to put into the Debian archives, so that users can install it with `apt install editorconfig-checker`. Since Ubuntu imports almost all packages from Debian, it will be possible on Ubuntu (and Mint and so on) as well, but only in the upcoming releases (hopefully this will land in Debian 11, maybe it will even be available in Ubuntu 20.10). I don't plan to provide a PPA though since I don't use Ubuntu, but someone that does can do it pretty easily based on the Debian package.

> Could you please add your change to the changelog? That would be great

Will do later.

stephanlachnit:
Pushed with changelog

PR #116, opened 2020-07-13
Ignore autogenerated *.yml.enc RoR credentials to prevent new line error

Mayurifag:


mstruebing:
Hey, at first: thank you very much for your contribution :rocket: 

Could you explain which tool these files generated?
What was your reason to add this to the core instead of ignoring the file for the test with the `exclude` option or to add a rule in the `.editorconfig` for that?

PR #128, opened 2020-10-19
Update max_line_length validator to work with UTF8, fixes #115

### LINKED ISSUE #115: max_line_length should support non ASCII characters
Currently it seems like `ä`, `ö` and `ü` are counted as two characters which makes the `max_line_length` rule not usable for not plain ASCII files (like UTF-8 so everything with emojis which maybe should also count as one character ;))

MasonM:
This fixes the `max_line_length` validator so it's charset-aware. For utf-8, it will now use the rune count in the string instead of the byte count, which fixes #115. Unfortunately, I couldn't figure out how to easily add support for UTF-16 because the "unicode/utf16" package doesn't provide a `RuneCountInString` method.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 ### Deprecated
 ### Removed
 ### Fixed
+* Fix `max_line_length` validation for UTF-8 [#115](https://github.com/editorconfig-checker/editorconfig-checker/issues/115)
 ### Security
 ### Misc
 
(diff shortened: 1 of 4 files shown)


mstruebing (commented):
Hey, again, great PR :+1: 
Could you also add that change to the `CHANGELOG.md` file?

MasonM:
@mstruebing  Okay, done in 0736bb93f21154100d894b9fb6daeef6ab49a62f

mstruebing (approved):

PR #132, opened 2020-10-28
Ignore patch files by default.

notpeter:
I know you may be hesitant to let the default exclude list grow, but patch files are common in repositories and definitely should never be checked for editorconfig violations.

I am aware that I can use .ecrc or CLI options to explicitly exclude them (which I have in my repos).

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -220,6 +220,7 @@ If you don't pass the `ignore-defaults` flag to the binary these files are exclu
 "\\.7z$",
 "\\.bz2$",
 "\\.log$",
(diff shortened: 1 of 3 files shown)


mstruebing:
Would you mind to open a pull request with that included? :)

notpeter:
> Would you mind to open a pull request with that included? :)

This is a PR ;)

mstruebing:
Sorry, my bad, I will release a new patch version soonish :+1:

PR #134, opened 2020-11-13
Instead of git ls-tree use filepath.Walk

tomtomsen:
Currently files must be added to git index to be verified.

I do not like the `git` dependency therefor removed it completely and used `filepath.Walk` instead.
What if i dont have git installed, what if i use a different version control system?

I remove the `git ls-tree` command and used `filepath.Walk` instead.
In my opinion its a breaking change, so be careful!

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -47,6 +47,8 @@ var defaultExcludes = []string{
 	"\\.js\\.map$",
 	"min\\.css$",
 	"min\\.js$",
+	"(^|/)\\.git($|/)",
+	"(^|/)node_modules($|/)",
 }
 
 var defaultAllowedContentTypes = []string{
--- a/pkg/config/config_test.go
+++ b/pkg/config/config_test.go
@@ -50,7 +50,7 @@ func TestGetExcludesAsRegularExpression(t *testing.T) {
 	}
 
 	actual = c.GetExcludesAsRegularExpression()
-	expected = `testfiles|yarn\.lock$|package-lock\.json|composer\.lock$|\.snap$|\.otf$|\.woff$|\.woff2$|\.eot$|\.ttf$|\.gif$|\.png$|\.jpg$|\.jpeg$|\.mp4$|\.wmv$|\.svg$|\.ico$|\.bak$|\.bin$|\.pdf$|\.zip$|\.gz$|\.tar$|\.7z$|\.bz2$|\.log$|\.patch$|\.css\.map$|\.js\.map$|min\.css$|min\.js$`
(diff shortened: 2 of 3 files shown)


mstruebing:
Hey, thank you for your PR, I agree that the git dependency it not ideal.

Could you do some measurements in a relatively large project and compare it with the current implementation?
I switched to git some time ago because I had performance issues in large directories, even with excluded files and hadn't another solution how to fix it.

mstruebing:
I did a quick comparison.

In a repository with: 81182 Files (including installed dependencies)
Your implementation took: `23.55s user 3.61s system 119% cpu 22.690 total`

Current master takes only:`3.24s user 0.28s system 115% cpu 3.048 total`

So it is a really large gap, and the repository I took would be even as a dependency in other projects, so it is not really feasible I think. Of course, it would be the right way to do it, but I think most use cases are a git repository anyway. 

If you are not in a git repository and say, i.e. `ec <directory>` it will check it without using `git ls-tree`

So: either we need a more speedy implementation (which I don't really know how to do) or leave it as it is.
Or do you have any other suggestions or viewpoints I do not see?
Thank you anyway :pray:

tomtomsen:
You're absolutely right, 24 seconds is way to slow. it seems, that filepath.Walk is known to be slow. 
i will look in some alternative libraries/solutions and check their performance. hope i can make it faster :)

currently my greatest pain-point is, that my pre-commit hook, does not validate newly added files and my ci/cd queue reports errors because of the those added (invalid formatted) files. if git is the (only fast) solution, i will try to add unstaged and staged files to the check.

mstruebing:
> if git is the (only fast) solution, i will try to add unstaged and staged files to the check

Sounds like a legit solution to me in case there isn't something else we could use.

I think it is slow because it will walk **any** file/directory and then check if it should be added. So every file needs to be processed somehow and with git we only process files which are tracked so we already eliminate dependencies here.

tomtomsen:
>  it will walk any file/directory
You are right, after adding a `filepath.SkipDir` if a directory is excluded, performance was way better. 
Then i needed to create an exclution expression, which was exactly the same as in the `.gitignore` file. which results in a exclusion redundancy which feels quite messy.

So i returned to my initial pain-point. As you mention before if `git` isn't available a normal `filepath.walk` will be executed which is fine. my pain-point was that newly added files which are currently not added to `git` repository or are currently staged files, are not getting checked before commit, therefor i added a lot of invalid files to my repository.

I looked closer into that problem and came up with a proper solution.

i will close this pull-request and create a new one.

Issue #99, opened 2019-12-19
Read excludes from `.gitignore` instead of using the git added files


mstruebing:
Currently we use `git ls-tree -r --name-only HEAD` to find files which should be checked.
It seems it would be better to read the `.gitignore` file, see: https://github.com/editorconfig-checker/editorconfig-checker.javascript/issues/111#issuecomment-567628706

mightyiam:
Thank you for considering this, @mstruebing.

Trying to imitate what git does in a project would probably be a bit more complicated than a single top-level `.gitignore`. If I remember correctly, any directory can have a `.gitignore` file and it gets rather complicated to re-implement all that.

Perhaps one of these could help, if it at all makes sense to execute git for this purpose.
https://git-scm.com/docs/git-check-ignore
https://git-scm.com/docs/git-ls-files

mstruebing:
fixed by #135

PR #135, opened 2020-11-16
Replace git ls-tree with git ls-files

tomtomsen:
`ls-tree` command returns all files in the git repository, ignoring all staged or newly added files.
`ls-files` adding those missing files.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@
 
 ### Fixed
 
+- Extended validation to staged and unadded files [#135](https://github.com/editorconfig-checker/editorconfig-checker/pull/135)
+
 ### Security
 
 ### Misc
--- a/pkg/files/files.go
+++ b/pkg/files/files.go
@@ -91,7 +91,7 @@ func GetFiles(config config.Config) ([]string, error) {
 		return filePaths, nil
 	}
 
-	byteArray, err := exec.Command("git", "ls-tree", "-r", "--name-only", "HEAD").Output()
+	byteArray, err := exec.Command("git", "ls-files", "--cached", "--others", "--modified", "--exclude-standard").Output()
 	if err != nil {
 		// It is not a git repository.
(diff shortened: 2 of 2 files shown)


mstruebing:
Hey thanks, that looks nice! :+1: 
Could you add that change to the `CHANGELOG.md` file?

mstruebing:
@tomtomsen Would you think this is a bugfix, feature or major release?
I'm not really sure

tomtomsen:
> @tomtomsen Would you think this is a bugfix, feature or major release?
> I'm not really sure

I don't think it's a major release, because BC is kept.
I don't think its a feature, because there is nothing new, its just "more".
So I'm tending to a bugfix
@mstruebing what do you think?

tomtomsen:
@mstruebing does this pull-request fix #99 ?

mstruebing:
> @mstruebing does this pull-request fix #99 ?

I think it is the same, yes.

One thing that came just now to my mind is:
If I have a file in my project which I do not track and do not want to track for some reasons (personal configuration of tools, not needed for the project/team members), this file will be checked too as we are now checking everything that is not ignored by git.
Could that leads to problems or is it okay to assume/force someone to format the files correctly when not checked in?

tomtomsen:
> One thing that came just now to my mind is:
> If I have a file in my project which I do not track and do not want to track for some reasons (personal configuration of tools, not needed for the project/team members), this file will be checked too as we are now checking everything that is not ignored by git.
> Could that leads to problems or is it okay to assume/force someone to format the files correctly when not checked in?
I don't think that will lead to problems. such files should, i my opinion be added to the project .gitignore, to a private user-.gitignore or in editorconfig-exclude pattern. worst case its well formatted.

lorengordon:
Just started using this tool, and was looking for exactly this feature. Any plans for a release soon?

mstruebing:
Thank you, I'm a bit busy right now, will try to do it as soon as possible :)

Issue #167, opened 2021-06-04
Document exclude regex is on relative path


polarathene:
I was using the `-dry-run` option which outputs absolute paths, and wondering why my config exclude was having trouble filtering out a top level directory...

Turned out that it's not applying regex based on the absolute path the dry run outputs, but [relative paths from where the command is run](https://github.com/editorconfig-checker/editorconfig-checker/pull/24)? That should be documented imo :sweat_smile:

mstruebing:
Thank you very much for that issue.
Feel free to provide a pull request regarding that documentation changes, that would be great 👍

PR #168, opened 2021-06-04
docs: Fix config example + note on regex use

### LINKED ISSUE #166: Docker image outputs blank version
`docker run --rm -it mstruebing/editorconfig-checker:2.3.5 ec -version` outputs nothing. I've also run the `ec -version` command via the containers shell internally and got the same result. Commands like `ec -help` work correctly.

This is an issue as even though the config specifies `2.3.5` in it's "Version" property, the tool exits stating versions do not match. I have seen in the source (which has some grammar and spelling errors in comment):

https://github.com/editorconfig-checker/editorconfig-checker/blob/8b1d55cc15bb55923f95208db1d80ab653fe5a26/cmd/editorconfig-checker/main.go#L17-L19

Is that meant to be [set in the Dockerfile](https://github.com/editorconfig-checker/editorconfig-checker/blob/master/Dockerfile)?

---

Additionally, in the README, you have two example configs, [one matches the source with PascalCase](https://github.com/editorconfig-checker/editorconfig-checker#configuration), while [the other is in snake_case](https://github.com/editorconfig-checker/editorconfig-checker#via-configuration). Did that convention change at some point and the snake_case example was not updated? Or are both styles supported? (not clarified in the README)

polarathene:
As a new user, an outdated config example and lack of clarification on filepaths for regex matching on tripped me up. Updated the README as requested to help others avoid those problems.

Fixes: #166 
Fixes: #167 

(Technically doesn't fix #166 , that still needs to build with the version included correctly into the Docker image. Resolves the other concern about documentation that was discussed)

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,9 +139,9 @@ A sample `.ecrc` file can look like this and will be used from your current work
     // set these options to true to disable specific checks
     "EndOfLine": false,
     "Indentation": false,
+    "IndentSize": false,
     "InsertFinalNewline": false,
     "TrimTrailingWhitespace": false,
-    "IndentSize": false,
     "MaxLineLength": false
   }
 }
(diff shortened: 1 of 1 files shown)


mstruebing (approved):
Thank you very much 🙏

Issue #170, opened 2021-06-07
Exclude .yarn/ and .pnp.cjs by Default


Kurt-von-Laven:
Yarn 2 promotes checking [dependencies](https://yarnpkg.com/features/zero-installs) along with [an auto-generated .pnp.cjs file](https://yarnpkg.com/features/pnp) into version control. By default editorconfig-checker tends to throw a lot of unhelpful errors on these files, so they should be excluded by default.

mstruebing:
Hey, thank you for that issue.
If you want, feel free to provide a pull request. 👍

Kurt-von-Laven:
I took a stab at it.

PR #171, opened 2021-06-08
feat: Exclude yarn 2 files by default (#170).

Kurt-von-Laven:
feat: Exclude yarn 2 files by default (#170). Add .yarn/, .pnp.cjs, and .pnp.js to the default excludes.

fix: Add $ to default exclude package-lock.json. This is more consistent with the rest of the default exclude patterns.


mstruebing (approved):
Thank you 🥇

PR #174, opened 2021-06-12
feat(exit-codes): return error on absent file

### LINKED ISSUE #157: Return error on absent file
It would be great to fail fast on bad command line input. When the file given is not there, there should not be a zero return code, but 1.

    $ ec no-such-file
    Could not get the ContentType of file: no-such-file
    stat no-such-file: no such file or directory
    
    $ echo $?
    0                   <------ should ideally be 1

    $ ec --version
    2.3.4

mstruebing:
Something is odd with the tests, need to investigate

closes #157


mstruebing:
Closing as I will not work on this PR anymore in the near future.

PR #183, opened 2021-10-12
Remove `git` in Dockerfile final stage

PeterDaveHello:
Looks like there is no need to have additional git dependency


PeterDaveHello:
@mstruebing oh, sorry that I didn't notice it, I only tested the binary directly on my machine, which already has git installed. Let me apologize and close this pull request. This is a cool tool by the way 😃

PR #195, opened 2022-02-17
Add `go install` method to README

silverwind:
I think this is a great way to install this tool on development and CI environments.


mstruebing:
Thank you 🚀

PR #202, opened 2022-06-13
feat: Exclude `.pnp.loader.mjs` by default (#201)

### LINKED ISSUE #201: Exclude `.pnp.loader.mjs` by Default
Yarn v3.1.0 generates a `.pnp.loader.mjs` file to implement ESM support for [Plug'n'Play](https://yarnpkg.com/features/pnp). This file is intended to be tracked in version control, but should be ignored by linters since it is auto-generated.

Kurt-von-Laven:
Fixes #201. This file is auto-generated by Yarn v3.1.0+.

fix: Add `^` to default excludes of specific files. Unlike most of the default excludes, which are for filetypes, `.yarn/`, `yarn.lock`, `package-lock.json`, `composer.lock`, `Cargo.lock`, `.pnp.cjs`, and `.pnp.js` are the full filenames of auto-generated files and folders. Hence, match the filenames exactly rather than excluding any filename ending in them.

docs: Trim trailing whitespace in README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -214,6 +214,7 @@ If you don't pass the `ignore-defaults` flag to the binary these files are exclu
 "^Cargo\\.lock$",
 "^\\.pnp\\.cjs$",
 "^\\.pnp\\.js$",
+"^\\.pnp\\.loader\\.mjs$",
 "\\.snap$",
 "\\.otf$",
 "\\.woff$",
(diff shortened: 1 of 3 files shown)


Kurt-von-Laven:
Thank you for creating such a helpful tool!

Issue #191, opened 2022-01-26
Errors in `.editorconfig-checker.json` are not reported and file is silently ignored


tudortimi:
I have a `.ecrc` file which contains an error: 

```
{
  "Version": "2.4.0",
  "Verbose": false,
  "Debug": false,
  "IgnoreDefaults": false,
  "SpacesAftertabs": false,
  "NoColor": false,
  "Exclude": ["\\.h$", "\\.c$",],
  "AllowedContentTypes": [],
  "PassedFiles": [],
  "Disable": {
    "EndOfLine": false,
    "Indentation": false,
    "InsertFinalNewline": false,
    "TrimTrailingWhitespace": false,
    "IndentSize": false,
    "MaxLineLength": false
  }
}
```

Notice the training comma after the final element of `Exclude`.

In this case, `ec` simply ignores the file and goes on to check everything. It would be helpful if an error message were printed.

tunetheweb:
Just ran into this, and it's particularly annoying because the comment in the [configuration docs](https://github.com/editorconfig-checker/editorconfig-checker#configuration) counts as making the whole file invalid:

https://github.com/editorconfig-checker/editorconfig-checker/blob/a6067702dd357ee10870df3b6734fd2e21a9045a/README.md#L149-L151

Probably should remove that at least as took me ages to figure out.

klaernie:
This was fixed with #380, which appeared in v3.1.0

PR #209, opened 2022-07-15
Remove breaking commit from example config file

tunetheweb:
Makes progress on #191 

It may nor be obvious to all (it wasn't to me) that the comment line MUST BE REMOVED or the whole config is invalid and is completely ignored.

Let's move it to a comment below.


mstruebing (approved):
Thank you for this contribution 🚀

PR #210, opened 2022-07-17
Fix ci

rasa:
Fixes CI failures. See https://github.com/editorconfig-checker/editorconfig-checker/pull/206#issuecomment-1186418624


mstruebing (approved):

.github/workflows/ci.yml
```diff
@@ -1,6 +1,8 @@
 name: "ci"
 
-on: [push]
+on:
+  push:
+  workflow_dispatch:
```
mstruebing, line 5: What does this do?

pkg/files/files.go
```diff
@@ -143,6 +143,7 @@ func ReadLines(content string) []string {
 
 // GetContentType returns the content type of a file
 func GetContentType(path string, config config.Config) (string, error) {
+
```
mstruebing, line 146: Could you remove this empty added line?

.github/workflows/ci.yml
rasa: That allows one to run the workflow manually. See [here](https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/).

pkg/files/files.go
rasa: Yeah, sorry. That got added by accident.

PR #214, opened 2022-07-27
Update .editorconfig

aaronchenwei:


aaronchenwei:
Sorry for any confusion. I wasn't aware which GH project page I stayed just now. I intended to modify file for another project.

PR #215, opened 2022-07-31
fix: add missing text content types to validator

rasa:
Somehow, I got these two lists out of sync. Would you be open to my rewriting this so there is just a single list, containing the regexs, and we strip out the metacharacters for
https://github.com/editorconfig-checker/editorconfig-checker/blob/ec3af1e38f494ed0ae96f9d0ee2d4d14f75d3f65/pkg/config/config.go#L109
?

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -56,6 +56,7 @@ var defaultExcludes = []string{
 	"min\\.js$",
 }
 
+// keep synced with pkg/validation/validation.go#L20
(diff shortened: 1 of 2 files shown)


mstruebing:
Sure yeah that would be a good refactor. I'm not sure ow hard it is to handle the metacharacters but if it is possible without much effort we should do it.

mstruebing (approved):

PR #216, opened 2022-08-03
fix: Fix typo in final EOL check error message

silverwind:
I think the intention of this error message is to show the absence of a final newline, so `no` makes much more sense than `new` to me.


mstruebing (approved):

mstruebing:
Thank you very much for your contribution 🚀

Issue #145, opened 2021-01-29
Support more common one line output format


lsouchard:
Though the output of editorconfig-checker is human friendly, it is complicated to parse in other tools. A more repetitive one line output format is basic but at least somewhat common. An option to fully print each error per line would be great.

As an example of integration, Mega-Linter is suggested but its integration in Github workflows doesn't suit me. I'd like to use [reviewdog](https://github.com/reviewdog/reviewdog) to report the errors provided by editorconfig-checker. However, the current output is not parseable by it. Reviewdog supports parsing from a given pattern as long as it is one error per line.

For now, I've created a small bash script to convert editorconfig-checker to a reviewdog compatible format ("%f:%l: %m"). I post it here in case someone needs it (to use it provide editorconfig-checker result in a pipe or pass a file containing it as argument) : 
```
#!/usr/bin/env bash

converted_lines=()
current_file=""
while IFS= read -r line; do
    if [[ $line =~ ^[0-9]+[[:blank:]]+errors[[:blank:]]+found$ ]]; then
        break
    elif [[ $line =~ ^[[:blank:]]*$ ]]; then
        continue
    elif [[ $line =~ ^[^[:blank:]]+:$ ]]; then
        current_file="$line"
    else
        if [[ $line =~ ^[[:blank:]]+[0-9]+: ]]; then
            converted_lines+=("$current_file$line")
        else
            converted_lines+=("${current_file}0:$line")
        fi
    fi
done < "${1:-/dev/stdin}"

printf '%s\n' "${converted_lines[@]}" | sed -E 's/:[[:blank:]]*([0-9]+):/:\1:/;s/(0:)[[:blank:]]+/\1 /'
```

mstruebing:
Either add a flag for `--human|--human-readable` and make machine readable output the default or add a flag like `--machine`? I don't know which would be better.

lsouchard:
I would argue for a flag to enable a machine-friendly output and not break editorconfig-checker's default output format. It would also make sense with the flag `--no-color` which is already implemented as a mod to the human-friendly default output.

Now a flag `--machine` may be a bit restrictive. I would rather have a flag like `--machine-format=<format>` taking as argument a pattern with flags to represent each data available to log (ex: the one provided in my example of reviewdog that are marked with a "%" and a character). This flag could also be used without an argument to apply a default one.

mstruebing:
That's a great idea! I like that. Thank you

dbast:
+1

[...]

PR #228, opened 2022-10-06
feat: Added functionality to output a more parsable format

timrydefalk:
Introduced a `-parsable` flag that enables outputting in similar format as linters like `flake8`, `mdl`, `shellcheck --format gcc` and `yamllint -f parsable` to name a few.


timrydefalk:
Should provide a solution for #145 

Let me know if I need to make any changes.

rasa (approved):

KaffeMyers:
Why was this never actually merged? :hamburger:

PR #247, opened 2023-01-21
docs: fix typo

hituzi-no-sippo:


mstruebing (commented):
Hey, thank you for your contribution and the pull request.
I think we should rather change this in the code to be `SpacesAfterTabs` as this is the more accurate name.
Would you be willing to provide a PR for this too?

hituzi-no-sippo:
Thank you for response.

> Hey, thank you for your contribution and the pull request.
I think we should rather change this in the code to be SpacesAfterTabs as this is the more accurate name.
Would you be willing to provide a PR for this too?

If you are okay with the following conditions, I will provide this PR:

* I am a beginner in Go language.
* I plan to create the PR starting in February.

If you would like, please answer the following a question:

* Is this PR Breaking change?

mstruebing:
In theory, it is a breaking change as this changes the API.
But as this is only the result of a typo, I would say it is a fix and would approve it as a fix commit.

hituzi-no-sippo:
Sorry for the late reply.

> Would you be willing to provide a PR for this too?

I had missed it.

I should just create a separate Pull Request for the breaking change to rename `SpacesAftertabs` to `SpacesAfterTabs`.

And should I close this Pull Request?

mstruebing:
I'll close it. Would be great if you could provide this change in a separate PR 👍

Issue #232, opened 2022-12-08
Confusion on official cli binary name


jay-babu:
Hello,

When this tool is installed, should the binary/cli tool be called `editorconfig-checker` or `ec`. Typically, shorthand names like `ec` are reserved for system tools such as `ls`, `cat`, etc.

When running

`go install github.com/editorconfig-checker/editorconfig-checker/cmd/editorconfig-checker@latest`, the binary name is `editorconfig-checker`. However, on the releases page, the binary is `ec`. This has sparked some discussions across a few plugins using this plugin:

https://github.com/jose-elias-alvarez/null-ls.nvim/pull/1282

https://github.com/jay-babu/mason-null-ls.nvim/issues/37

https://github.com/williamboman/mason.nvim/pull/722

What should these plugins assume the binary to be called by default: `editorconfig-checker` or `ec`?

jay-babu:
Note: I am a fan of `editorconfig-checker`, but don't let that change what it is supposed to be.

jay-babu:
@chalkygames123, would you be able to provide some guidance?

chalkygames123:
Nah, I'm not a maintainer of this project. But I'm a fan of `editorconfig-checker` too.

jay-babu:
@mstruebing, are you a maintainer? Would you be able to weigh in on this?

mstruebing:
So, my intention to add `ec` as a possible binary name was just because it is shorter and I'm lazy!
I did not really know about the convention of system tools.

Do you have a strong opinion if it should rather be ec or editorconfig-checker?

@chalkygames123 @jay-babu 

I currently think it should be `editorconfig-checker` but I let your opinion count more than mine.

jay-babu:
I would side with `editorconfig-checker`, based on the above convention. People usually alias if they want something shorter. It's a 1 line change for them.

chalkygames123:
Having an alias binary is okay, but having **only** that one is not a good idea, IMO.

jay-babu:
@mstruebing, can we close this issue with the acceptance of `editorconfig-checker` as the resolution?

jemag:
If truly the confusion is to be removed, then I think it would be important to standardize it across the board.

Currently the `pacman` install and the `docker` image use `ec` and the README is also consistently referring to `ec`. For those reasons, I actually thought that the official binary name was `ec`.

 If `editorconfig-checker` is to be adopted instead, then it would be important to change all of those.

[...]

PR #255, opened 2023-02-15
chore!: rename to editorconfig-checker

jay-babu:


jay-babu:
Closes: https://github.com/editorconfig-checker/editorconfig-checker/issues/232

theoludwig (changes requested):
Thank you for the PR! :smile:

If we rename the command to `editorconfig-checker` instead of `ec`, we should also replace all occurrences of `ec` in the README.

klaernie:
hm.. that change in README.md would only be two lines - Should I rebase @jay-babu's commit, add that small change and open a new PR?

klaernie:
done in #371

PR #303, opened 2023-10-11
feat: enable cached parser

greut:
Reading and parsing the ini file is done *once* per ini file rather than *the number of .editorconfig files* per file. So O(n) -> O(1) when it comes to reading the Editorconfig definition.

It doesn't seem to affect much the performance on the quick benchmark I did on saltstack/salt.


mstruebing:
@greut the tests are failing - can you figure out if it has to do with your changes or something else is up?
You can execute them by running `make test`.

PR #319, opened 2023-12-28
feat: add output format option with 'gcc'

vnayar:
Relating to #145, a new command-line option  was added. The current output format is named 'default', and a new format 'gcc' was added that uses an error format similar
to gcc, namely:  `<file>:<lineNo>:<charNo>: <type/severity>: <message>`

This command-line option mirrors that used by the common linting tool shellcheck: https://www.mankier.com/1/shellcheck

Example output:
```
$ ./bin/ec --format=gcc ~/files/dummy.*
../../files/dummy.bat:0:0: error: Wrong line endings or no final newline
../../files/dummy.bat:0:0: error: Not all lines have the correct end of line character
../../files/dummy.d:2:0: error: Line too long (113 instead of 100)
../../files/dummy.d:3:0: error: Wrong amount of left-padding spaces(want multiple of 2)
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,8 @@ USAGE:
         show which files would be checked
   -exclude string
         a regex which files should be excluded from checking - needs to be a valid regular expression
+  -format
+        specifies the output format, see "Formats" below for more information
   -h    print the help
   -help
         print the help
@@ -131,6 +133,14 @@ USAGE:
 If you run this tool from a repository root it will check all files which are added to the git repository and are text files. If the tool isn't able to determine a file type it will be added to be checked too.
 
 If you run this tool from a normal directory it will check all files which are text files. If the tool isn't able to determine a file type it will be added to be checked too.
+
+### Formats
+
+The following output formats are supported:
+
+- **default**: Plain text, human readable output.
+- **gcc**: GCC compatible output. Useful for editors that support compiling and showing syntax errors.
+           <file>:<line>:<column>: <type>: <message>
 
 ## Configuration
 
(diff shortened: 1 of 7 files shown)


vnayar:
I modified the commit description to have the prefix `feat:`, which I believe is what was causing the build to fail, although I'm not 100% sure.  Please approve execution of the workflow.

vnayar:
Another force-push due to the subject line of the commit being rejected with the following error:
```
✖   subject must not be sentence-case, start-case, pascal-case, upper-case [subject-case]
```
I am guessing that capitalizing the letters in the subject will resolve this, but I have no way of checking until the workflow is run.

vnayar:
I cannot find any documentation regarding the proper case for commit message titles, but all the other examples that are present use lower case. I am making yet another attempt to satisfy this commit linting tool. Please re-run the workflow, because currently only maintainers can do this.

vnayar:
Updated `testfiles/generated-config` to include an empty "format" value.

vnayar:
When time permits, please take a look. The option to alter the output format is called "--format" and the first added alternative is "gcc". The output format and commnd-line options are inspired from https://www.shellcheck.net/

vnayar:
@mstruebing Could you take a moment to look at this PR to add a `--format` option, with one additional format so far, `--format=gcc`, which outputs line-oriented output similar to `gcc`. The option name and value were borrowed from the common linting tool `shellcheck`.

mstruebing (approved):
Thank you this is a good PR.
Sorry for taking so long.

mstruebing:
I've created a new release, 2.8.
Can #145 be closed or does this issue still needs to be open or can it be closed?

vnayar:
> I've created a new release, 2.8. Can #145 be closed or does this issue still needs to be open or can it be closed?

I believe all the described conditions of #145 have been satisfied, namely:
> Though the output of editorconfig-checker is human friendly, it is complicated to parse in other tools. A more repetitive one line output format is basic but at least somewhat common. An option to fully print each error per line would be great.

PR #324, opened 2024-02-01
feat: add .pnm, .pbm, .pgm and .ppm files extensions to default excludes

theoludwig:
The PR adds more images formats to exclude by default.

You can read more about these formats here: https://netpbm.sourceforge.net/doc/pnm.html.

As they are binary files (if not using the plain version), they should be ignored.


mstruebing (approved):
LGTM Thank you

PR #339, opened 2024-04-18
fix: v3 prefix is required

### LINKED ISSUE #338: Unable to `go run` v3 because module version does not tag version
The go module version of the v3 tags is still v2, making it impossible to use v3 via `go run` and likely also `go install`:

```bash
$ go run github.com/editorconfig-checker/editorconfig-checker/v2/cmd/editorconfig-checker@v3 -help
go: github.com/editorconfig-checker/editorconfig-checker/v2/cmd/editorconfig-checker@v3.0.1: github.com/editorconfig-checker/editorconfig-checker/v2@v3.0.1: invalid version: module path includes a major version suffix, so major version must match
```

BTW I totally blame this on golang which makes v2+ modules [so hard](https://github.com/golang/go/issues/31543).

greut:
Since that package is now a v3, it needs to be reflected in the `go.mod` file.

Closes #338


theoludwig (approved):

theoludwig:
Thanks!

mstruebing (approved):
❤️

Issue #245, opened 2023-01-17
More Output Formats


steffenf14:
Hi, 
are there plans to add more output formats like junit?
My goal is to have a better integration in gitlab pipelines for example.
I have seen such implementations in deptrac or php codesniffer

mstruebing:
I did not plan it yet.
If someone is willing to provide a PR I will sure have a look at it and merge it if it's good.

jerone:
Issue #317 requests an output for GitHub Actions.

jerone:
Besides the mentioned formats (junit & GitHub Actions), I would like to ask for API to define **custom formats** myself.

I'm working in a project with multiple linters and I'm writing custom formatters/outputs/reporters for them to return an constant output logging.
So far I was able to write reporters for [CSpell](https://cspell.org/types/cspell-types/interfaces/CSpellReporter.html) & [ESLint](https://eslint.org/docs/latest/extend/custom-formatters).

KaffeMyers:
I think the format that is used by flake8 (and a lot of other linters) should be one of the standard options, as it is an established output standard used by several linters and is easy to parse:
``` shell
$ flake8
./job-generator.py:1:5: E116 unexpected indentation (comment)
```

PR #360, opened 2024-08-15
feat: consolidate adjacent error messages

### LINKED ISSUE #359: it would be nice if editorconfig-checker could collect instances of the same issues found in adjacent lines into a single report line
I've just put a bunch of generated json documents into one of our repositories, but forgot to either ignore them or adjust their indentation to fit our editorconfig.

However this lead to me getting flooded with about 10k lines of reports that the indentation was incorrect.

I think a better output format would collect consecutive issues by not outputting each line by line, but rather turning consecutive lines having the same issue into a range. Although I can imagine that this naive approach would fail in cases where each line has two errors reported (they would ping-pong, interrupting each other's consecutiveness).

klaernie:
This changes the human output format to collect multiple identical error
messages on consecutive lines into one error message that list both the
starting and ending line.

This is still missing a properly written testcase, I so far only tested
with

    make
    (cd testfiles; ../bin/ec multiple-wrong-indentations.txt)

So far the output for the newly included testfile is this:

    multiple-wrong-indentations.txt:
            2-4: Wrong amount of left-padding spaces(want multiple of 10)
            8-9: Wrong amount of left-padding spaces(want multiple of 10)
    6 errors found
    1 files checked

However there are still a few things that annoy me:
1. The summary of how many errors are present is now incorrect, since the `error.GetErrorCount()` is called by `main()`, before any consolidation happened.
However one should not hoist the consolidation up into `main()`, since it must only happen for the human output (or the eventual github action format mentioned in #245).
My first instinct would be to let `error.PrintErrors()` handle writing that summary as well, but technically that would add a side effect for `error.TestPrintErrors()`.

2. I have no idea how to work with golang's testing infrastructure. I'd roughly guess that one would have to make PrintErrors() side effect free by collecting the lines to be printed and returning them, letting the caller handle the actual printing, but being then able to validate the output lines in a testcase.

I'd be really happy for any feedback, while programming for years, this is maybe my third time touching golang - which is quite a different beast compared to Perl. So especially if any I did is not idiomatic or hard to understand, please tell me! (aka: `perlcritic --brutal`)


mstruebing:
This looks like a great PR already.
I will need to find some time to look over the changes and think about it.
If someone other maintainer has time to have a look that would be very much appreciated 🫶 

CI is failing because of your commit message, we require semantic commits i.e. `feat: blabla`. 
I'm not entirely sure we use that still anywhere, so we might be able to remove that check but if we want to move further with automatic releasing in the future it could help to have semantic commits.

klaernie:
that's a really easy fix at least - I fixed the commit messages, treating the change to the human output as a breaking change, since somebody out there is probably going to rely on it. Sadly I didn't see a CONTRIBUTING.md or similar section in the README and didn't notice in the existing commits.

Since it didn't seem much work anymore, I also implemented the output format that github actions uses.

pkg/error/error.go
```diff
@@ -10,8 +10,9 @@ import (
 
 // ValidationError represents one validation error
 type ValidationError struct {
-	LineNumber int
-	Message    error
+	LineNumber        int
+	Message           error
+	ConsecuitiveCount int
```
ccoVeille, line 15:
```suggestion
	ConsecutiveCount int
```

pkg/error/error.go
```diff
@@ -20,6 +21,13 @@ type ValidationErrors struct {
 	Errors   []ValidationError
 }
 
+func (error1 *ValidationError) Equal(error2 ValidationError) bool {
+	return error1.Message.Error() == error2.Message.Error() &&
+		error1.LineNumber == error2.LineNumber &&
+		error1.ConsecuitiveCount == error2.ConsecuitiveCount
```
ccoVeille, line 27:
```suggestion
		error1.ConsecutiveCount == error2.ConsecutiveCount
```

pkg/error/error.go
```diff
@@ -31,6 +39,44 @@ func GetErrorCount(errors []ValidationErrors) int {
...
+
+	// scan through the errors
+	for i := 0; i < len(errorsWithLines); i++ {
+		thisError := errorsWithLines[i]
+		config.Logger.Debug("investigating error %d(%s)", i, thisError.Message)
+		// scan through the errors after the current one
+		for j, nextError := range errorsWithLines[i+1:] {
+			config.Logger.Debug("comparing against error %d(%s)", j, nextError.Message)
+			if nextError.Message.Error() == thisError.Message.Error() && nextError.LineNumber == thisError.LineNumber+thisError.ConsecuitiveCount+1 {
+				thisError.ConsecuitiveCount++ // keep track of how many consecutive lines we've seen
```
ccoVeille, line 67:
```suggestion
			if nextError.Message.Error() == thisError.Message.Error() && nextError.LineNumber == thisError.LineNumber+thisError.ConsecutiveCount+1 {
				thisError.ConsecutiveCount++ // keep track of how many consecutive lines we've seen
```

pkg/error/error.go
```diff
@@ -51,12 +102,29 @@ func PrintErrors(errors []ValidationErrors, config config.Config) {
...
+						config.Logger.Error(fmt.Sprintf("::error file=%s::%s", relativeFilePath, singleError.Message))
+					}
+				}
 			} else {
 				// default: A human readable text format.
 				config.Logger.Warning(fmt.Sprintf("%s:", relativeFilePath))
 				for _, singleError := range fileErrors.Errors {
 					if singleError.LineNumber != -1 {
-						config.Logger.Error(fmt.Sprintf("\t%d: %s", singleError.LineNumber, singleError.Message))
+						if singleError.ConsecuitiveCount != 0 {
```
ccoVeille, line 123:
```suggestion
						if singleError.ConsecutiveCount != 0 {
```

pkg/error/error.go
```diff
@@ -51,12 +102,29 @@ func PrintErrors(errors []ValidationErrors, config config.Config) {
...
+							config.Logger.Error(fmt.Sprintf("\t%d-%d: %s", singleError.LineNumber, singleError.LineNumber+singleError.ConsecuitiveCount, singleError.Message))
```
ccoVeille, line 124:
```suggestion
							config.Logger.Error(fmt.Sprintf("\t%d-%d: %s", singleError.LineNumber, singleError.LineNumber+singleError.ConsecutiveCount, singleError.Message)),
```

klaernie:
I now managed to build a test case for the newly introduced `error.ConsolidateErrors()` function - hopefully next time Codecov runs this should bring the test coverage back up again.

ccoVeille (changes requested):
Typo spotted

klaernie:
of cause now testing the tester is needed as well - that finally should bring testing coverage up to where it was before I started this branch

klaernie:
okay, now the ci should pass - turns out I had a unicode nonbreaking space in the second to last commit, and the wording in the last commit tripped the "subject must not be sentence-case" rule. Now I verified locally to ensure the commit messages pass.

pkg/error/error_test.go
```diff
@@ -112,6 +113,69 @@ func TestGetErrorCount(t *testing.T) {
...
+}
+
+func TestConsolidateErrors(t *testing.T) {
+	input := []ValidationError{
+		// two messages that become one block
+		{LineNumber: 1, Message: errors.New("message kind one")},
+		{LineNumber: 2, Message: errors.New("message kind one")},
+		// one message with a good line between it and the last bad line, but repeating the message
+		{LineNumber: 4, Message: errors.New("message kind one")},
+		// one message that breaks the continousnes of line 4 to line 6
```
ccoVeille, line 155:
```suggestion
		// one message that breaks the continuousness of line 4 to line 6
```

pkg/error/error_test.go
```diff
@@ -112,6 +113,69 @@ func TestGetErrorCount(t *testing.T) {
...
+		{LineNumber: 5, Message: errors.New("message kind two")},
+		{LineNumber: 6, Message: errors.New("message kind one")},
+		// one message without a line number, that will become sorted to the top
+		{LineNumber: -1, Message: errors.New("file-level error")},
+	}
+
+	expected := []ValidationError{
+		{LineNumber: -1, Message: errors.New("file-level error")},
+		{LineNumber: 1, ConsecutiveCount: 1, Message: errors.New("message kind one")},
```
ccoVeille, line 164: I know lists start at 0

But I would have expected to find a Consecutive count of 2.

With the current example and code, line 1 will report a consecutive count of 1.

A human would expect to know that the error reported line 1 was reported twice, so 2

pkg/error/error_test.go
```diff
@@ -112,6 +113,69 @@ func TestGetErrorCount(t *testing.T) {
...
+	}
+	if baseError.Equal(differentMessageError) {
+		t.Error("failed to detect a difference in the Message")
+	}
+	if baseError.Equal(differentCountError) {
+		t.Error("failed to detect a difference in the ConsequtiveCount")
+	}
+}
+
+func TestConsolidateErrors(t *testing.T) {
```
ccoVeille, line 148: I would like to have one more tests

About this example

Line 1: error1
Line 2: error2
Line 2: error1
Line 3: error3
Line 3: error1
Line 3: error2
Line 4: error4

I would expect to have:
- error1 detected 3 times starting at line 1
- error2 detected 2 times starting at line 2
- error3
- error4

I'm unsure about what could render the current implementation. And a test like this is important

pkg/error/error_test.go
klaernie: hm.. ConsecutiveCount is the count of additional error messages that followed the first one - so it excludes the first one (meaning `ConsecutiveCount == 0` is "one singular message").
But switching to storing according to your model would mean the last-line-of-block calculation becomes a tad simpler. I'll think about that one - maybe renaming the property to a better name is  better to make it clearer (I'm thinking about `AdditionalIdenticalErrorCount` for example)

ccoVeille: It would make sense.

I'm not good with wording, but yes I think it would make it easier to understand.

pkg/error/error_test.go
klaernie: This interleaved pattern is definitely not supported by the current implementation - it would return the list exactly the same as it was put it. As it is currently the implementation is basically a run-length-encoding compression of the error message, nothing more.

Supporting that would require keeping track of multiple different errors - which I'm not confident I could easily do, given my skill level in golang. In my mind that would require first converting the flat list of errors into a map per line, keyed by the error, then scanning that list again to RLE each error type, than merging these results again into a flatter list by sorting by starting line number and error message. I think that would be more than it needs to be, comparing it to the current state running in production.

pkg/error/error_test.go
klaernie: thanks, fixed.

pkg/error/error_test.go
klaernie: I renamed it, sadly I had no better idea than `AdditionalIdenticalErrorCount`, so that one it is for  now ;)

ccoVeille: Someone may come and suggest a new name

pkg/error/error_test.go
ccoVeille: I can understand that.

But I'm afraid, this would be expected.

Said otherwise, I think people will open issues about it.

So it's up to maintainers to choose whether they are OK to merge.

I might be able to help with the implementation in a few days. But I cannot for now, as I only have my phone with me.

Maybe one thing to do is to add the test and use a `t.Skip` to have the test ready for a future refactoring.

ccoVeille:
For anyone interested by the feature this PR would bring, please take a look at

https://github.com/editorconfig-checker/editorconfig-checker/pull/360#discussion_r1720906778

pkg/error/error_test.go
klaernie: I'm open for suggestions - a rename is cheap!

testfiles/multiple-wrong-indentations.txt
theoludwig, line 1: Is this file used?
I think, the  `testfiles` are used for test cases in https://github.com/editorconfig-checker/editorconfig-checker/blob/main/pkg/validation/validation_test.go but I don't think it is applicable here for your feature.

theoludwig (commented):
Thank you @klaernie for opening the PR, and thank you @ccoVeille for helping to review the PR. :smile:

I've left some comments.

cmd/editorconfig-checker/main.go
```diff
@@ -39,8 +39,8 @@ func init() {
 	flag.BoolVar(&c.ShowVersion, "version", false, "print the version number")
 	flag.BoolVar(&c.Help, "help", false, "print the help")
 	flag.BoolVar(&c.Help, "h", false, "print the help")
-	flag.StringVar(&c.Format, "format", "default", "specify the output format: default, gcc")
-	flag.StringVar(&c.Format, "f", "default", "specify the output format: default, gcc")
+	flag.StringVar(&c.Format, "format", "default", "specify the output format: default, gcc, github-actions")
+	flag.StringVar(&c.Format, "f", "default", "specify the output format: default, gcc, github-actions")
```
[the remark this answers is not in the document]
klaernie: `github-actions` is the format explicitly used by Github Actions, actually a separate output format that just was easy to implement once the consolidation was implemented.

The consecutive grouping is done in both the `default` and `github-actions` output formats, so all users will get this.

testfiles/multiple-wrong-indentations.txt
klaernie: You are indeed correct. My first understanding lead me to testing this way, but this in no way is needed going forward. I've removed it.

testfiles/.editorconfig
```diff
@@ -21,3 +21,7 @@ max_line_length = 2
 [spaces-with-tab.c]
 indent_style = space
 indent_size = unset
+
+[multiple-wrong-indentations.txt]
+indent_style = space
+indent_size = 10
```
theoludwig, line 27: As the "testfile"  has been removed in [`42f72b3` (#360)](https://github.com/editorconfig-checker/editorconfig-checker/pull/360/commits/42f72b341487979acfb5e6a36f9cbdb3d712f843), those lines should also be removed.

cmd/editorconfig-checker/main.go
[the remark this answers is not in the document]
theoludwig: I'm not sure to understand. Can you show a difference of output between the `default` and `github-actions` format? Is it a conventional format used by GitHub Actions, if so is there any documentation from GitHub?

That might be out of the scope of this PR (can be done separately) but it seems like the `PrintErrors` function is not covered by tests yet? Could be a good idea to cover this code by snapshot tests, to also easily see the difference of outputs between the formats available.

theoludwig (approved):

pkg/error/error_test.go
ccoVeille: Did you add such a test for future refactoring?

testfiles/.editorconfig
klaernie: oh, good point. Thanks, I totally forgot that one. Now it is removed.

cmd/editorconfig-checker/main.go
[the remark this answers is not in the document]
klaernie: > That might be out of the scope of this PR (can be done separately) but it seems like the `PrintErrors` function is not covered by tests yet? Could be a good idea to cover this code by snapshot tests, to also easily see the difference of outputs between the formats available.

I discovered the untestedness / untestability of `PrintErrors` while opening the PR. I'm mainly unsure how to approach it - one might argue that `PrintErrors()` should become `RenderErrors()`, also include outputting the summary as I mentioned in list item 1 in [my initial PR comment](https://github.com/editorconfig-checker/editorconfig-checker/pull/360#issue-2468383591), and simply return the formatted text (as one big string).

Then testing this new `RenderErrors()` function would become trivial, and the actual output with `config.Logger.Error()` would need to be still done by https://github.com/klaernie/editorconfig-checker/blob/consolidate/cmd/editorconfig-checker/main.go#L127-L130, and the error package would be side effect free (except for mutating the `ValidationErrors.Errors` array).

pkg/error/error_test.go
klaernie: I only came around to doing this just now, but in commit 5a402006d92321a73a19df04714aa07e4f7f5e00 I tried roughly representing the example you gave. I also left a warning for anybody implementing this about the assumption I made.

cmd/editorconfig-checker/main.go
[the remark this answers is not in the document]
klaernie: I followed through with testing the `PrintErrors`  function, although it now is called `FormatErrors`.

theoludwig:
@klaernie, I've created an issue about the maintenance of `editorconfig-checker`, if you would like more context. :smile:

https://github.com/editorconfig-checker/editorconfig-checker/issues/364

klaernie:
Having seen the latest CI run result, I'm not sure what the best way to proceed with snapshot testing is.

The test cases in [error_test.go line 229](https://github.com/klaernie/editorconfig-checker/blob/6198132bd3286890458ce3ab7880f633c47c1ccb/pkg/error/error_test.go#L229) and [error_test.go line 238](https://github.com/klaernie/editorconfig-checker/blob/6198132bd3286890458ce3ab7880f633c47c1ccb/pkg/error/error_test.go#L238) refer to files by an absolute patch, and the [conversion to a relative path](https://github.com/klaernie/editorconfig-checker/blob/6198132bd3286890458ce3ab7880f633c47c1ccb/pkg/error/error.go#L86) then means the test case becomes sensitive to the current working directory. If a user tries to run the test in `/home/user/ecc/pkg/error` they will get a different result compared to if their current directory is `/home/user/three/more/dirs/ecc/pkg/error`.

The only workaround I can imagine is changing the working directory for the duration of the test - which I implemented in 0d679db54110f8e69fbf989b4a52d04325ea832c. If anyone has a better solution feel free to tell me.

pkg/error/error.go
```diff
@@ -20,6 +22,13 @@ type ValidationErrors struct {
 	Errors   []ValidationError
 }
 
+func (error1 *ValidationError) Equal(error2 ValidationError) bool {
+	return error1.Message.Error() == error2.Message.Error() &&
+		error1.LineNumber == error2.LineNumber &&
+		error1.AdditionalIdenticalErrorCount == error2.AdditionalIdenticalErrorCount
+
+}
```
ccoVeille, line 30: This code might be refactored (it's up to you)

To something like this,making it more readable 


```suggestion
func (error1 *ValidationError) Equal(error2 ValidationError) bool {
	if error1.LineNumber != error2.LineNumber {
		return false
	}

	if error1.Message.Error() != error2.Message.Error() {
		return false
	}

	return true
}
```

I inverted the test on message and line, comparing integers is faster

I removed the test on count, because I don't think you [need](https://github.com/editorconfig-checker/editorconfig-checker/pull/360#discussion_r1739717042) it

pkg/error/error.go
```diff
@@ -20,6 +22,13 @@ type ValidationErrors struct {
 	Errors   []ValidationError
 }
 
+func (error1 *ValidationError) Equal(error2 ValidationError) bool {
+	return error1.Message.Error() == error2.Message.Error() &&
+		error1.LineNumber == error2.LineNumber &&
+		error1.AdditionalIdenticalErrorCount == error2.AdditionalIdenticalErrorCount
```
ccoVeille, line 28: I don't get the test on the additional identical count

Could you explain?

pkg/error/error.go
klaernie: I didn't know that early exits are idiomatic golang - but I definitely like them to make code more readable. I'll implement that!

pkg/error/error.go
klaernie: Since `Equal()` is used in the test case, it needs to verify all properties - else once might have a test case succeed that just reports the first error and does not report a correct ending number. This method is not used for the actual consolidation - it would have to be another method `Similar()` (which than would need to ignore the `AdditionalIdenticalErrorCount`

klaernie:
One thing I just remembered, that I need to do: adjust the README. Please hold me accountable.

pkg/error/error.go
klaernie: rewritten in e16817eefdda4c9a425931d2b47f53f030a36fed

pkg/error/error.go
ccoVeille: Ok it makes sense now

ccoVeille (approved):

pkg/error/error.go
ccoVeille: Idiomatic or not the readability always should come first.

A long line with 3 conditions is never a good thing 😅😂

klaernie:
Dotting `i`s and crossing `t`s I noticed one problem with the snapshot test: they fail under Windows:

```text
kandre@Windows10 MINGW64 ~/git-repos/editorconfig-checker (consolidate)
$ go test ./...
go: downloading github.com/editorconfig/editorconfig-core-go/v2 v2.6.2
go: downloading github.com/gabriel-vasile/mimetype v1.4.5
go: downloading github.com/gkampitakis/go-snaps v0.5.7
go: downloading github.com/baulk/chardet v0.1.0
go: downloading golang.org/x/text v0.17.0
go: downloading golang.org/x/mod v0.17.0
go: downloading gopkg.in/ini.v1 v1.67.0
go: downloading github.com/gkampitakis/ciinfo v0.3.0
go: downloading github.com/gkampitakis/go-diff v1.3.2
go: downloading github.com/kr/pretty v0.3.1
go: downloading github.com/maruel/natural v1.1.1
go: downloading github.com/tidwall/gjson v1.17.0
go: downloading github.com/tidwall/pretty v1.2.1
go: downloading github.com/tidwall/sjson v1.2.5
go: downloading github.com/kr/text v0.2.0
go: downloading github.com/rogpeppe/go-internal v1.12.0
go: downloading github.com/tidwall/match v1.1.1
go: downloading golang.org/x/net v0.27.0
?       github.com/editorconfig-checker/editorconfig-checker/v3/cmd/editorconfig-checker        [no test files]
ok      github.com/editorconfig-checker/editorconfig-checker/v3/pkg/config      0.386s
ok      github.com/editorconfig-checker/editorconfig-checker/v3/pkg/encoding    2.988s
--- FAIL: TestFormatErrors (0.02s)
    error_test.go:298:
        - Snapshot - 0
        + Received + 2

        @@ -1,7 +1,7 @@

          []logger.LogMessage{
        -     {Level:"warning", Message:"proc/cpuinfo:"},
        +     {Level:"warning", Message:"/proc/cpuinfo:"},
              {Level:"error", Message:"\t1: WRONG"},
        -     {Level:"warning", Message:"proc/cpuinfoNOT:"},
        +     {Level:"warning", Message:"/proc/cpuinfoNOT:"},
              {Level:"error", Message:"\t1: WRONG"},
              {Level:"warning", Message:"some/other/path:"},
              {Level:"error", Message:"\tWRONG"},

        at __snapshots__\error_test.snap:2

    error_test.go:301:
        - Snapshot - 2
        + Received + 2

        @@ -1,6 +1,6 @@

          []logger.LogMessage{
        -     {Level:"error", Message:"proc/cpuinfo:1:0: error: WRONG"},
        -     {Level:"error", Message:"proc/cpuinfoNOT:1:0: error: WRONG"},
        +     {Level:"error", Message:"/proc/cpuinfo:1:0: error: WRONG"},
        +     {Level:"error", Message:"/proc/cpuinfoNOT:1:0: error: WRONG"},
              {Level:"error", Message:"some/other/path:1:0: error: WRONG"},
              {Level:"error", Message:"some/other/path:0:0: error: WRONG"},
              {Level:"error", Message:"some/file/with/consecutive/errors:1:0: error: message kind one"},

        at __snapshots__\error_test.snap:20

    error_test.go:304:
        - Snapshot - 2
        + Received + 2

        @@ -1,6 +1,6 @@

          []logger.LogMessage{
        -     {Level:"error", Message:"::error file=proc/cpuinfo,line=1::WRONG"},
        -     {Level:"error", Message:"::error file=proc/cpuinfoNOT,line=1::WRONG"},
        +     {Level:"error", Message:"::error file=/proc/cpuinfo,line=1::WRONG"},
        +     {Level:"error", Message:"::error file=/proc/cpuinfoNOT,line=1::WRONG"},
              {Level:"error", Message:"::error file=some/other/path::WRONG"},
              {Level:"error", Message:"::error file=some/other/path,line=1::WRONG"},
              {Level:"error", Message:"::error file=some/file/with/consecutive/errors::file-level error"},

        at __snapshots__\error_test.snap:35

FAIL
FAIL    github.com/editorconfig-checker/editorconfig-checker/v3/pkg/error       0.374s
ok      github.com/editorconfig-checker/editorconfig-checker/v3/pkg/files       2.779s
ok      github.com/editorconfig-checker/editorconfig-checker/v3/pkg/logger      0.312s
ok      github.com/editorconfig-checker/editorconfig-checker/v3/pkg/utils       0.312s
ok      github.com/editorconfig-checker/editorconfig-checker/v3/pkg/validation  0.427s
ok      github.com/editorconfig-checker/editorconfig-checker/v3/pkg/validation/validators       0.247s
FAIL

kandre@Windows10 MINGW64 ~/git-repos/editorconfig-checker (consolidate)
```

Thinking about it we have three options:
1. drop the snapshot tests entirely
2. skip the snapshot tests `if os.PathSeparator != "/"`
3. differentiate the snapshots by `os.PathSeparator`, so that the stored snapshots are different for each variant.

I would categorically exclude option 1, since we do not give up that easily.

Option 2 is a per-OS version of options 1, preventing future develops who work on Windows from benefiting from the snapshot tests, but would at least codify that the current tests are only expected to work on systems with unix-style path separators.

Option 3 would be probably the best, but introduce a new long-term problem: running the testsuite fully would require being done twice - once on Windows and once on Linux. Running in CI or Linux one can at least install wine and a windows version of golang, then run `wine go.exe test ./...`. Not running the testsuite on Windows would risk getting the snapshots on windows out of sync with the linux version - which could be a tricky surprise in the future.

Does anyone of you have a suggestion what route to take?

theoludwig:
Ah, Windows and path issues. 🙃

Hmm, we should probably run CI in Windows and macOS as well. For the moment, only `ubuntu-latest` is tested with GitHub Actions.

Given, that we run CI only on Linux, we should probably do:
> skip the snapshot tests if os.PathSeparator != "/"

Unless we already have a solution, we might create an issue to fix that separately to this PR. Run CI in all supported OS.

ccoVeille:
I would say as long as ubuntu test are launched via the CI, and as long as it's a know and documented things, I'm fine with tests that use t.Skip when `os.PathSeparator != "/"`

and if a better solution is found later, then it will be fixed

But yes, windows, and macOS should be added to the matrix

klaernie:
Okay, if we take the long term plan of running CI on Windows and MacOS as well, option three (using separate tests per separator) would be probably the best. I'll then make sure we programmatically run the same input into all of the tests, so they are guaranteed to be equal in terms of content, but make sure they use separate snapshots per path separators.
Essentially it's two sets of linked tests, where one of them is skipped each run, depending on the OS. 
Then in the future CI testing both platforms will guarantee that no outdated snapshots remain for long.

klaernie:
I've now completed enabling the tests to work with both path separators golang supports, and ensured that a possible future discovery of new path separator fails the test.

I would now call this finally complete enough to merge.

pkg/error/error_test.go
ccoVeille: I think we should open an issue about the improvement suggestion that is now in a skipped test

klaernie: Done: #373

xsy420:
I tried to build this package with this PKGBUILD and this pr (since this pr to be precise)
```shell
pkgname=editorconfig-checker-git
_pkgname=editorconfig-checker
_commit=cf4ae1ccede331b2aa1b115f1de5257737de7eef
pkgver=next
pkgrel=1
pkgdesc='A tool to verify that your files are in harmony with your .editorconfig'
arch=(x86_64)
url="https://github.com/editorconfig-checker/$_pkgname"
license=(MIT)
depends=(glibc)
makedepends=(go)
provides=(ec)
conflicts=("$_pkgname")
source=(
        "git+${url}.git#commit=$_commit"
)
sha256sums=('SKIP')

build() {
        cd "$srcdir/$_pkgname"
        export CGO_CPPFLAGS="$CPPFLAGS"
        export CGO_CFLAGS="$CFLAGS"
        export CGO_CXXFLAGS="$CXXFLAGS"
        export CGO_LDFLAGS="$LDFLAGS"
        export GOFLAGS="-buildmode=pie -trimpath -ldflags=-linkmode=external -mod=readonly -modcacherw" # won't pass check
        # export GOFLAGS="-buildmode=pie -ldflags=-linkmode=external -mod=readonly -modcacherw" # will pass check
        go build -o "$_pkgname" ./cmd/...
}

check() {
        cd "$srcdir/$_pkgname"
        go test ./...
}

package() {
        cd "$srcdir/$_pkgname"
        install -Dm755 -t "$pkgdir/usr/bin/" "$_pkgname"
        ln -sf "$_pkgname" "$pkgdir/usr/bin/ec"
        install -Dm644 -t "$pkgdir/usr/share/licenses/$_pkgname" LICENSE
}

```
I got this error
```
==> Checking runtime dependencies...
==> Checking buildtime dependencies...
==> Retrieving sources...
  -> Updating editorconfig-checker git repo...
==> Validating source files with sha256sums...
    editorconfig-checker ... Skipped
==> Extracting sources...
  -> Creating working copy of editorconfig-checker git repo...
Cloning into 'editorconfig-checker'...
done.
Switched to a new branch 'makepkg'
==> Starting build()...
==> Starting check()...
?       github.com/editorconfig-checker/editorconfig-checker/v3/cmd/editorconfig-checker        [no test files]
ok      github.com/editorconfig-checker/editorconfig-checker/v3/pkg/config      0.004s
ok      github.com/editorconfig-checker/editorconfig-checker/v3/pkg/encoding    0.034s
--- FAIL: TestFormatErrors (0.00s)
    error_test.go:320: mkdir github.com: permission denied
    error_test.go:323: mkdir github.com: permission denied
    error_test.go:326: mkdir github.com: permission denied
FAIL

Snapshot Summary

✕ 3 snapshots failed

FAIL    github.com/editorconfig-checker/editorconfig-checker/v3/pkg/error       0.003s
ok      github.com/editorconfig-checker/editorconfig-checker/v3/pkg/files       0.009s
ok      github.com/editorconfig-checker/editorconfig-checker/v3/pkg/logger      (cached)
ok      github.com/editorconfig-checker/editorconfig-checker/v3/pkg/utils       0.002s
ok      github.com/editorconfig-checker/editorconfig-checker/v3/pkg/validation  0.017s
ok      github.com/editorconfig-checker/editorconfig-checker/v3/pkg/validation/validators       (cached)
FAIL
==> ERROR: A failure occurred in check().
    Aborting...
```

If I remove `-trimpath` in `GOFLAGS`, it can build fine. Is it ok if I (or next archlinux official package) just remove this flag?

mstruebing:
@orhun you are maintaining the package, could you have a look?

orhun:
That flag is [necessary](https://wiki.archlinux.org/title/Go_package_guidelines#Flag_meaning) for the reproducibility. I just checked and the package builds fine in a [clean chroot](https://wiki.archlinux.org/title/DeveloperWiki:Building_in_a_clean_chroot) - you can remove the flag if it causes problems in your local machine :)

orhun:
You need to build in a clean chroot with `extra-x86_64-build` command instead of running the build steps manually.
cc @alerque for more input on this.

klaernie:
@xsy420, @orhun: I broke this conversation into issue #397, since the reason behind the build failure is more difficult, and not related to the build environment at all.

Issue #321, opened 2024-01-22
Improved compatibility with Windows by providing zip binary archives


Okeanos:
As very briefly mentioned in #204

> Then, I think it's better to have Windows releases in zip format, since there is no .tar.gz support natively, I use 7zip for that.

It would be very, very cool if you could provide Windows binary releases as `.zip` archive instead of `.tar.gz` – Windows isn't natively able to unpack/work with `.tar.gz`. This makes it especially troublesome because tools such as `winget` are also [not](https://github.com/microsoft/winget-cli/issues/2899) able to deal with `.tar.gz` either.

Providing `.zip` files would open up a lot of additional fancy integration possibilities via e.g. `winget` and related tooling making adoption on Windows way easier and nicer.

Okeanos:
Closed by accident when going through my list of open issues. Apologies for any confusion.

Okeanos:
@mstruebing I may be able to provide a PR for this but just to be sure I am not overlooking anything too obvious … the relevant change would be in the [Makefile](https://github.com/editorconfig-checker/editorconfig-checker/blob/main/Makefile#L155-L159) and probably involve a Windows vs. everything else check, right?

mstruebing:
@Okeanos yes you are right. Except you would want to compress all binaries as zips.
The wrapper (php, javascript, pyhton) will need to adjust anyway so I would think one format which can be used anywhere would be better than making exceptions - if that's possible. But I guess zip is common on every OS.

theoludwig:
This issue should probably be closed as https://github.com/editorconfig-checker/editorconfig-checker/issues/362 have been merged?

PR #362, opened 2024-08-21
feat: add zip version when compressing all binaries (#321)

Okeanos:
This adds a zip archive to the output of the compress-all-binaries sub-task in order to provide better compatiblitiy with Windows.

Windows does not come with the ability to handle tar.gz (until now the only output format) and, hence, users cannot easily use the prebuilt binaries distributed via GitHub. Further integration into, for example, the winget community packages is also prevented by this.

Zip was chosen as an _additional_ output format in order to not create an unexpected/breaking change for consumers of the existing tar.gz format.

Windows is not handled specifically here in order to keep the process of providing build artifacts as simple as possible.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -155,6 +155,7 @@ _build-all-binaries:
 _compress-all-binaries:
 	for f in bin/*; do      \
 		tar czf $$f.tar.gz $$f;    \
+		zip $$f.zip $$f;    \
 		rm -f $$f;    \
 	done
 


theoludwig (approved):
Thank you for your contribution! :smile:

I agree with this comment: https://github.com/editorconfig-checker/editorconfig-checker/issues/321#issuecomment-1968457907
> Except you would want to compress all binaries as zips.
The wrapper (php, javascript, pyhton) will need to adjust anyway so I would think one format which can be used anywhere would be better than making exceptions - if that's possible. But I guess zip is common on every OS.
---

> Zip was chosen as an additional output format in order to not create an unexpected/breaking change

I think in that case the breaking change might be worth it, to not increase too much the size of our release assets. And the breaking change should be pretty much invisible for end users, when the wrappers will be updated.

What do you think? @mstruebing

Okeanos:
> And the breaking change should be pretty much invisible for end users, […].

Don't count on it: I am currently consuming the `tar.gz` release file directly from within some GitHub Actions because for a brief period the v3 release broke the `go run` syntax (fixed in https://github.com/editorconfig-checker/editorconfig-checker/pull/339) and I never bothered to switch back.

I can obviously adjust this but having my workflows break unexpectedly with no prior notice (again) would be somewhat of a bummer. I would kindly ask that at least for 1-2 releases both archives will be published with a proper notification in the release notes.

mstruebing:
In all honesty, I think the wrappers should be deprecated.
I'm not entirely sure how widely they are used so it's hard to decide.
They add unneeded complexity to the process and the tool which is just not worth it and can be error prone.
The tool should be easily installable and usable and we should support for major ways, i.e. github actions and nix are coming to my mind. We should not create a wrapper for every programming language just to be able to use the programming languages package manager. This was a mistake by me a long time ago. What I kind of would like/support is a clear seperation between the main tool and the wrappers (a separate GH org) that is related but not officially supported/maintained but rather community organised.

Additionally, the whole release should be reworked altogether, see https://github.com/editorconfig-checker/editorconfig-checker/issues/204 which is quite old already.

Sorry for the long rant :)

Now to the answer:
I guess we can just publish `tar.gz` and `zip` files, we could remove the `tar.gz` in a future release and, if needed/wanted, adjust the wrappers to download the `zip` file, which should also be supported and preinstalled on most operating systems.

mstruebing (approved):
Thank you 🚀 , I will leave it open for 1-2 days to see if there are more responses

theoludwig (approved):

polarathene:
> remove the `tar.gz` in a future release

Generally this is more accessible for linux, similar to zip for Windows.

`tar` to archive a directory of files followed by a compression format like gzip. There is no `zip` equivalent there, you'd sometimes need to install the `unzip` package IIRC which in some environments (at least with containers) is more common to not have that package vs `tar` and `gzip`/`xz`.

I'm not sure what's common on macOS.

---

Other feedback:
- I've not looked into the archive contents, but if it were just a static binary you could probably just offer that without the compression step? 🤔 
- I assume the binary is static linked since there's no distinction of `glibc`/`gnu` vs `musl` on the linux ones that I'm used to seeing.
- It's good that the version name is not in the file name as tends to only have a disadvantage (_you couldn't rely on the `latest` release if you weren't version pinning for some reason_).
- There is quite a lot of niche platforms being published that's probably not needed if you're concerned about total size of all assets published for a release. If you're going to change expectations that require workflows to adjust beyond a version bump, then you might as well drop all those uncommon ones and only publish them if an actual request is made to keep publishing the artifacts for those niche platforms?

PR #363, opened 2024-08-23
feat: editorconfig-checker-disable-next-line

### LINKED ISSUE #180: feat: editorconfig-checker-disable-next-line
It may seen redundant to introduce such a feature at first, but here's [GNU Make](https://www.gnu.org/software/make/manual/html_node/Makefile-Contents.html)

> You cannot use comments within variable references or function calls: any instance of # will be treated literally (rather than as the start of a comment) inside a variable reference or function call.

I bet there are other instances out on the wild, that cannot do with a inline comment to disable editorconfig-checker for that specific line.

ccoVeille:
- **fix: broken test**
- **chore: fix typos and style in documentation**
- **feat: add editorconfig-checker-next-line**

Fixes #180

### DIFF
--- a/pkg/validation/validation.go
+++ b/pkg/validation/validation.go
@@ -106,7 +106,7 @@ func ValidateFile(filePath string, config config.Config) []error.ValidationError
 
 		// check for the status of the previous line (it was the next line on previous loop iteration)
 		if disableNextLineFound {
-			// editorconfig-checker-disable--nextline was found on previous line
+			// editorconfig-checker-disable-next-line was found on previous line
 
 			// check for successive editorconfig-checker-disable-next-line
 			disableNextLineFound = strings.Contains(line, directiveDisableNextLine)
--- a/pkg/validation/validation.go
+++ b/pkg/validation/validation.go
@@ -33,6 +33,7 @@ func ValidateFile(filePath string, config config.Config) []error.ValidationError
 	const directiveDisable = directivePrefix + "disable"
 	const directiveDisableFile = directivePrefix + "disable-file"
 	const directiveDisableLine = directivePrefix + "disable-line"
+	const directiveDisableNextLine = directivePrefix + "disable-next-line"
 	const directiveEnable = directivePrefix + "enable"
 
 	var validationErrors []error.ValidationError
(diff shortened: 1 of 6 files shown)


ccoVeille:
please review @theoludwig @mstruebing 

cc @andreineculau if you are still interested as you created the issue #180 3 years ago
cc @BorzdeG because we talked about it with https://github.com/caarlos0/env/pull/327

pkg/validation/validation.go
```diff
@@ -95,22 +96,60 @@ func ValidateFile(filePath string, config config.Config) []error.ValidationError
...
-				// found editorconfig-checker-disable-line, so just skip current line
-				continue
-			} else {
-				isDisabled = true
-			}
+		}
+
+		// check for the status of the previous line (it was the next line on previous loop iteration)
+		if disableNextLineFound {
+			// editorconfig-checker-disable--nextline was found on previous line
```
ccoVeille, line 109:
```suggestion
			// editorconfig-checker-disable-next-line was found on previous line
```

mstruebing:
Amazing 🚀 
I will probably review this PR this weekend.

testfiles/disabled-line.txt
```diff
@@ -1,3 +1,3 @@
 hello
-	world // editorconfig-checker-disable-line
+  world // editorconfig-checker-disable-line
```
theoludwig, line 2: Why the indentation (from tabs to spaces) changed here?

theoludwig (commented):
Thank you, looks mostly good to me. :+1:

testfiles/disabled-line.txt
ccoVeille: Please look at the first commit.

Let me know if I have to explain more

theoludwig: Yes, I saw. But this change is unrelated to the feature this PR is trying to add, and could be made in a separated PR. 
I tried both with and without this change, and tests are passing both ways, as you explained it is because the code was already valid, so there was nothing (no errors) to ignore.

What confused me is that this PR is changing behavior close to `editorconfig-checker-disable-line` implementation, which made me think, `editorconfig-checker-disable-line` doesn't work anymore as intended.

theoludwig (approved):

testfiles/disabled-line.txt
ccoVeille: I can split the PR, no problem. We can talk about each of them separately.

theoludwig:
> Are we good then?
> 
> I wasn't sure with the last messages from @theoludwig

@ccoVeille To me it looks good, ready to merge as is, without further changes.
**BUT** I'm really only reviewing general logic and unit tests, and if the behavior matches with expectations (as a heavy user of `editorconfig-checker` myself).
I'm not coding in Go (actually never used it), and don't own this codebase or did much in it, so I can't and should not have the confidence to review implementation details.

---

For more context about the maintenance of `editorconfig-checker`, I've created an issue: https://github.com/editorconfig-checker/editorconfig-checker/issues/364

ccoVeille:
OK, thanks. Feel free to merge whenever you want then

testfiles/disabled-line.txt
ccoVeille: A reply was provided here

https://github.com/editorconfig-checker/editorconfig-checker/pull/363#issuecomment-2322607033

It's OK

ccoVeille:
Quick follow-up

The directive was used twice in 3 months

https://github.com/search?q=%22editorconfig-checker-disable-next-line%22+-is%3Afork+-org%3Aeditorconfig-checker&type=code

I feel proud 😅

Issue #365, opened 2024-09-04
Provide code quality report as JSON according to Code Climate specificiation


tploss:
For projects hosted on GitLab the following feature would probably help many developers to better integrate editorconfig-checker in their GitLab CI flow.

Coming from GitLab and reading [its documentation on integrating a custom tool for code quality](https://docs.gitlab.com/ee/ci/testing/code_quality.html#implement-a-custom-tool), we should be able to provide a code quality report as JSON according to [Code Climate specificiation](https://github.com/codeclimate/platform/blob/master/spec/analyzers/SPEC.md#data-types).

As documented in the first link, the Code Quality report artifact JSON file must contain an array of objects with the following properties:

Name | Description
-- | --
description | A description of the code quality violation.
fingerprint | A unique fingerprint to identify the code quality violation. For example, an MD5 hash.
severity | A severity string (can be info, minor, major, critical, or blocker).
location.path | The relative path to the file containing the code quality violation.
location.lines.begin | The line on which the code quality violation occurred.

This would be an alternative to the already existing gcc and default format.
As requested in #245 I'd be willing to implement this myself and open a PR.

klaernie:
interesting, that #367 did not close this issue.

ccoVeille:
> interesting, that merging #367 did not close this issue.

The PR was mentioning "more details #id",
not "Fix #id", or "Closes #id"

It seems "normal"

PR #367, opened 2024-09-05
feat: provide Codeclimate compatible report fromat

tploss:
Provide code quality report as JSON according to Code Climate specificiation. This is need when using the editorconfig-checker in GitLab CI.

More details: #365


klaernie (changes requested):
Looks generally good and like it adheres to the specification you left in the comment. Thanks for turning that ugly `if...elseif..else` into a `switch` - I totally forgot to think about if that feature is available in golang when writing it.. ;)

However I'd like to see tests for demonstrating it in action and proving correctness.

Furthermore I think you should disable the verbose and debug logging when `config.Format == codeclimate` - else anyone trying to debug will surround the JSON output with unparsable noise.

Also I thought about, if one should maybe turn the codeclimate generation into a method of `ValidationErrors`, which only runs `ConsolidateErrors` and returns the JSON.

In that regard, I was wondering lately, if it would be better to split `FormatErrors` into per-output-format functions, and change their interface towards  `main()` to utilize an io.Writer. But that could easily be another PR once this one is merged.

pkg/error/format_codeclimate.go
```diff
@@ -0,0 +1,50 @@
...
+	return CodeclimateIssue{
+		Check:       checkName,
+		Description: err.Message.Error(),
+		Fingerprint: fingerprint,
+		Severity:    severity,
+		Location: CodeclimateLocation{
+			Path: path,
+			Lines: CodeclimateLines{
+				Begin: err.LineNumber,
+				End:   err.LineNumber, // Currently we only have one line per issue
```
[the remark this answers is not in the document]
tploss: Totally overlooked this when adapting for the new consolidated lines. Thanks 😃

tploss:
> Looks generally good and like it adheres to the specification you left in the comment. Thanks for turning that ugly `if...elseif..else` into a `switch` - I totally forgot to think about if that feature is available in golang when writing it.. ;)
> 
> However I'd like to see tests for demonstrating it in action and proving correctness.

👍 
Will definitely create some tests. Just have to first understand this *snapshot* system for test data.

> Furthermore I think you should disable the verbose and debug logging when `config.Format == codeclimate` - else anyone trying to debug will surround the JSON output with unparsable noise.

An easier solution from my perspective would be to just log non-standard messages (so debug, error, etc) on stderr instead of stdout. Then one can easily just use `editorconfig-checker --format codeclimate > report.json` but still see all error and debug messages.
Thoughts?

> Also I thought about, if one should maybe turn the codeclimate generation into a method of `ValidationErrors`, which only runs `ConsolidateErrors` and returns the JSON.
> 
> In that regard, I was wondering lately, if it would be better to split `FormatErrors` into per-output-format functions, and change their interface towards `main()` to utilize an io.Writer. But that could easily be another PR once this one is merged.

I was also not really happy with how a single function would grow per added format but did not want to refactor that completely in my PR that just wants to add a single format.
Can we do that as a separate PR afterwards?

klaernie:
> Will definitely create some tests. Just have to first understand this _snapshot_ system for test data.

It's actually fairly easy to use: you just demonstrate calling your code in error_test.go and hand the result of calling `FormatErrors` to `s.MatchSnapshot`. Then run `UPDATE_SNAPS=true go test ./...` so it generates the new snapshot.

tploss:
> It's actually fairly easy to use: you just demonstrate calling your code in error_test.go and hand the result of calling `FormatErrors` to `s.MatchSnapshot`. Then run `UPDATE_SNAPS=true go test ./...` so it generates the new snapshot.

Done 🙂 
I also verified the test by changing the severity constant to "major" and this subsequently failed the test case.

tploss:
The ci is waiting for an approval before running??
I'm not really used to GitHub CI so I don't know if this is normal. @klaernie @mstruebing

klaernie:
Indeed, the CI is waiting to be approved by a maintainer of the repo (so sadly I cannot trigger it).

klaernie (approved):

klaernie:
Oh, just noticed you overlooked the same places in the README and cmd/main.go argument help as me, which show the available output formats.

mstruebing:
CI already ran so someone needs to have pressed the button.
It's a mechanism to make sure someone who does a PR doesn't break anything or leak any secrets etc.
You could do all kind of bad stuff with Github Actions, this makes sure that at least someone needs to approve the run.

tploss:
> Oh, just noticed you overlooked the same places in the README and cmd/main.go argument help as me, which show the available output formats.

Thanks, I have fixed it and also provided a nicely formatted example output 🙂

cmd/editorconfig-checker/main.go
```diff
@@ -127,7 +127,9 @@ func main() {
 	errorCount := len(formattedErrors)
 	if errorCount != 0 {
 		config.Logger.PrintLogMessages(formattedErrors)
-		config.Logger.Error(fmt.Sprintf("\n%d errors found", errorCount))
+		if config.Format != "codeclimate" {
```
ccoVeille, line 130: I would add a comment to explain why you disable the log

ccoVeille, line 130: I would like to suggest adding a constant for code climate

This way code that relies on this value could be easily identified

ccoVeille (approved):

cmd/editorconfig-checker/main.go
klaernie: I would take it on me to split `FormatErrors` into multiple independent functions once this PR is merged. Now with a fourth format supported this evergrowing function is too hard to read.

I also wondered if there was a neat way to represent the collection of possible output formats, so that the argument parsing code can discover the possible formats without us having to every time adjust it by hand.

ccoVeille: I would love that

It would help to have a type and a method to list all the formats.

The fact this code has to be maintained is q bit a pain

```go
        flag.StringVar(&c.Format, "format", "default", "specify the output format: default, gcc, github-actions, codeclimate")
```

klaernie: Opened #374 to remind me of the refactoring. 

Reading up on golang, I somehow did not find any mention how to implement essentially an enum type. @ccoVeille do you have seen anything alike in the past?

theoludwig:
I've merged this PR, as it seems like it was ready to go. :smile:

Feel free to open follow up PRs for refactoring.

PR #371, opened 2024-09-06
chore!: rename ec to editorconfig-checker

### LINKED ISSUE #232: Confusion on official cli binary name

klaernie:
This is the replacement for #255 

# Still TODO:
- [x] replace image in README.md
- [x] add support for the renamed binary in all downstream consumers


mstruebing (approved):
This will be a hard breaking change.
I will leave it open for other maintainers to have a look at.

Once we are fine with it and we want to release a new version I can go ahead an do the release until the release process is sorted out.

klaernie:
Of cause! Please tell me if I missed any spot, I tried to be as thorough as possible, but could still have missed a spot.

theoludwig (changes requested):
**BREAKING CHANGE** indeed, might be worth it, to avoid confusions, and avoid having too many way to execute the same binary.

What about `.ecrc`? Should the configuration file named `.editorconfig-checkerrc`? (might be a separate issue, but as it's a JSON file, we should support the file with the extension, so `.editorconfig-checkerrc.json`).
**EDIT:** Issue already open for JSON config: https://github.com/editorconfig-checker/editorconfig-checker/issues/306

Also, wrappers will need to be updated:
- https://github.com/editorconfig-checker/editorconfig-checker.javascript/pull/416
- https://github.com/editorconfig-checker/editorconfig-checker.python/pull/33
- https://github.com/editorconfig-checker/editorconfig-checker.php/pull/321

We had issues with releasing major version with Go from v2 to v3 (ref: https://github.com/editorconfig-checker/editorconfig-checker/issues/338), so we should not forgot to update it the same way as https://github.com/editorconfig-checker/editorconfig-checker/pull/339 did for v4 (of course in a separated PR).

We might first release v3.1.0 (hopefully with a automation), and then we'll see how to proceed with major version.

Blocking as "Request changes" for the moment.

ccoVeille:
> What about .ecrc? Should the configuration file named .editorconfig-checkerrc? 

I commented #306 about this point.

https://github.com/editorconfig-checker/editorconfig-checker/issues/306#issuecomment-2335034329

But let me sum up here, I think this PR is a good candidate for checking a new name in addition of .ecrc file

I suggest `.editorconfig-checker.json` (not `.editorconfig-checkerrc.json`as suggested here, there is no reason for me to keep the `rc` suffix)

It might not need to be addressed in this exact PR, but in another PR that will be added to the same release that where the binary name changes would be shipped)

Dockerfile
```diff
@@ -6,7 +6,7 @@ COPY . /ec
 ARG TARGETOS TARGETARCH
 RUN --mount=type=cache,target=/root/.cache/go-build \
     --mount=type=cache,target=/go/pkg \
-    GOOS=${TARGETOS} GOARCH=${TARGETARCH} GO111MODULE=on CGO_ENABLED=0 go build -ldflags "-X main.version=$(cat VERSION | tr -d '\n')" -o bin/ec ./cmd/editorconfig-checker/main.go
+    GOOS=${TARGETOS} GOARCH=${TARGETARCH} GO111MODULE=on CGO_ENABLED=0 go build -ldflags "-X main.version=$(cat VERSION | tr -d '\n')" -o bin/editorconfig-checker ./cmd/editorconfig-checker/main.go
```
polarathene, line 9: `GO111MODULE=on` is no longer relevant since Go 1.17:

```suggestion
    GOOS=${TARGETOS} GOARCH=${TARGETARCH} CGO_ENABLED=0 go build -ldflags "-X main.version=$(cat VERSION | tr -d '\n')" -o bin/editorconfig-checker ./cmd/editorconfig-checker/main.go
```

klaernie: I would not do this within the scope of this PR. We could remove relict that well before this breaking change gets merged.

polarathene: Yeah no worries, I just noticed it when glancing over the diff and thought I'd bring it to your attention 😅 (_Go 1.17 was released in Aug 2021, so it's been irrelevant for a while_)

theoludwig:
Do you think, it's still worth it to do the breaking change? @klaernie We might not need to release a v4.0.0 for a long long time, or maybe we do it now?

theoludwig (approved):
Finally time.

I will take care of updating the https://github.com/editorconfig-checker/editorconfig-checker.javascript wrapper, about the others I don't know.

Maybe to avoid being too breaking/disruptive, in the wrappers, we can try to first check for the `editorconfig-checker` binary, and then if doesn't work, as a fallback still use `ec`.

klaernie:
> Maybe to avoid being too breaking/disruptive, in the wrappers, we can try to first check for the `editorconfig-checker` binary, and then if doesn't work, as a fallback still use `ec`.

My first instinct was to version-gate this, but simply checking if there is a binary is also a way to approach it. But I think the main logic happens in picking the right archive to download from the release, which then determines the binary name.

theoludwig:
> replace image in README.md

@klaernie If you want, you can use this picture:

image

klaernie:
> > replace image in README.md
> 
> @klaernie If you want, you can use this picture:

Thanks, I was wondering if I'd spend the time to configure a terminal profile for the same look and feel or just pick something random.

mstruebing:
Something random would be fine, I'm not attached to that color scheme (and changed my personal one a lot of times in between).

klaernie:
> Something random would be fine, I'm not attached to that color scheme (and changed my personal one a lot of times in between).

I'm amazed now precisely @theoludwig replicated it already.

---

I think this PR is now at the point where we should merge it.

Issue #370, opened 2024-09-05
(Default) excludes and regex behavior


Okeanos:
I recently went through the file based excludes again for a project and when reading through the [default excludes](https://github.com/editorconfig-checker/editorconfig-checker?tab=readme-ov-file#default-excludes) I may have (originally) misunderstood some things.

The regex matching behavior isn't explicitly described and an exclusion such as `^composer\\.lock$` appears to tell editorconfig-checker to ignore a file that is named _exactly_ `composer.lock`. The exclusion `\\.jpg$` is supposed to match all files having a `.jpg` extension and `min\\.js$` matches anything ending in `min.js`, right?

Based on that understanding I want to ignore files called `.terraform.lock.hcl` and initially used the exclude pattern `^\\.terraform\\.lock\\.hcl` for that.

To my surprise the following happened:

```output
$ cat .ecrc
{
    "exclude": [
        "^\\.terraform\\.lock\\.hcl$"
    ]
}
$ editorconfig-checker
modules/repositories/.terraform.lock.hcl:
	5: Wrong indentation type(spaces instead of tabs)
```

When I now switch the regex to 

```output
$ cat .ecrc
{
    "exclude": [
        "\\.terraform\\.lock\\.hcl$"
    ]
}
```

editorconfig-checker successfully runs without finding any validation errors. What am I doing wrong? From what I can tell the `^...$` regex has a "problem" where it only matches files in the scanned root directory, not any subfolders. A look at the [`IsExcluded`](https://github.com/editorconfig-checker/editorconfig-checker/blob/main/pkg/files/files.go#L35-L51) function appears to confirm that where the regex test is used with a relative path inside the scanned root directory and not the filename.

Am I understanding all of this correctly and if so, would it make sense to:

- describe that behavior in the docs (with some examples) in detail?
- change some of the default excludes (starting with `^`) to also apply in subfolders out of the box (I am thinking about multi-module/monorepo structures)?

klaernie:
Your observations are totally correct - you copied the `composer.lock` exclusion rule, not being aware that the regex is applied to paths and not filenames. This of cause is confusing. I hope my rewording in #372 makes this much more easy to grasp.

PR #372, opened 2024-09-06
docs: try to explain excludes better for #370

klaernie:
Hopefully this provides some better explanations for the issue seen in #370.

Reading through the section in the README I found it confusing to have the general explanation at the far bottom, when it is applicable to the understanding of all the options.

I'm very aware that if this PR is merged before #371 I'll have to rebase that PR (or vice versa).

### DIFF
--- a/README.md
+++ b/README.md
@@ -354,7 +354,7 @@ An `.ecrc` which would ignore all test files and all Markdown files can look lik
 
 ##### via arguments
 
-If you want to play around how the tool would behave you can also pass the `--exclude` argument to the binary. This will accept a regular expression as well. The argument given will be added to the excludes as defined by your `.ecrc` (respecting both its `Exclude` and `IgnoreDefaults` settings).
+If you want to play around how the tool would behave you can also pass the `--exclude` argument to the binary. This will accept a regular expression as well. The argument given will be added to the excludes as defined by your `.ecrc` (respecting both its [`Exclude`](#via-configuration) and [`IgnoreDefaults`](#ignoring-default-excludes) settings).
 
 For example: `ec --exclude node_modules`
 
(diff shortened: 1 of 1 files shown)


klaernie:
I was also wondering if the options should be renamed:
- `IgnoreDefaults` to `IgnoreDefaultExcludePaths`
- `Exclude` to `ExcludePaths`

These might be added in addition to the existing names, being merged in. Technically that would then allow `.ecrc` files with both `.Exclude` and `.ExcludePaths`, but would retain support for all the `.ecrc` files currently out in the wild, while still making new configs better to read.

mstruebing (approved):
Thank you very much, I like the explanation.

README.md
```diff
@@ -229,7 +228,18 @@ const myTemplateString = `
...
 
-### Excluding Files
+### Excluding Paths
+
+You can exclude paths from being checked in several ways:
+- ignoring a file by documenting it inside the to-be-excluded file
+- adding a regex matching the path to the `.ecrc`
+- passing a regex matching the path as argument to `--exclude`
+
+All these excludes are used in addition to the [default excludes](#default-excludes), unless you opt out of them.
```
theoludwig, line 238: To make it clearer. :smile:

```suggestion
All these excludes are used in addition to the [default excludes](#default-excludes), unless you opt out of them with the `--ignore-defaults` flag.
```

theoludwig (approved):
Apart from the little suggestion, it looks good to me, thanks!

README.md
klaernie: I fixed this a slightly different way: I created an entire new section describing the `IgnoreDefaults` option and command line switch, linking to it from all the places it is mentioned.

PR #375, opened 2024-09-08
feat: support `.editorconfig-checker.json` config

### LINKED ISSUE #306: Support for .ecrc.json

theoludwig:
Fixes #306


theoludwig:
Feel free to review also @klaernie :smile:

cmd/editorconfig-checker/main.go
```diff
@@ -18,8 +18,8 @@ import (
 // version is dynamically set at compiletime
 var version string
 
-// defaultConfigFilePath determines where the config is located
-const defaultConfigFilePath = ".ecrc"
+// defaultConfigFileNames determines the file names where the config is located
+var defaultConfigFileNames = []string{".ecrc", ".ecrc.json", ".editorconfig-checkerrc", ".editorconfig-checkerrc.json"}
```
theoludwig, line 22: For the moment, we keep support for `.ecrc`, `.ecrc.json`, but when #367 lands, we will drop support for them.

README.md
```diff
@@ -150,7 +150,12 @@ The following output formats are supported:
 
 ## Configuration
 
-The configuration is done via arguments or an `.ecrc` file.
+The configuration is done via arguments or it will take the first config file found with the following file names:
+
+- `.ecrc`
+- `.ecrc.json`
+- `.editorconfig-checkerrc`
+- `.editorconfig-checkerrc.json`
```
theoludwig, line 158: I'd like to keep the `rc` from the filename, because it is a known convention in Unix/Linux, and many other tools also follow this convention.

ccoVeille: There is no use for so much format

ccoVeille: The rc suffix seems odd to me once you add an extension.

Also, you are adding too much formats, these one should be enough

- `.ecrc`
- `.editorconfig-checker.json`

Here is what I would suggest, and I don't think there is a need for the second one.

I could accept supporting

- `.ecrc`
- `.editorconfig-checkerrc.json`

But then this is it.

Then about the .json5 extension, klaernie suggested using .json5 extension

cmd/editorconfig-checker/main.go
ccoVeille: Was .edrc.json really supported ?

README.md
klaernie: I sadly know a couple dyslexic colleague who will fail to notice the doubled letter r in `checkerrc`. While your of cause correct that many tools use that convention, they equally support the convention that binary names should be as short as possible - trading readability for brevity. And this is exactly the pattern we're currently fixing here. Plus I just had to actually Google what `rc` stands for - and sadly "runtime configuration" is not what I found.

In my mind `.editorconfig-checker.json` is readable and understandable. But that is of cause only my opinion, and I'm not willing to die on a hill for it ;)

klaernie: > Then about the .json5 extension, klaernie suggested using .json5 extension

I did, but am becoming a bit wary of it - it requires us to replace a parser provided by the standard library with one only found in external modules, forcing us to make a choice between the implementations. Also some voices on the Internet argue against JSON5 and for other competing standards trying to achieve the same goals. So I'm internally divided on what the best way is.

cmd/editorconfig-checker/main.go
theoludwig: No, `.ecrc.json` wasn't supported, but as we're adding support for `.editorconfig-checkerrc.json`, it might makes sense to also support `.ecrc.json`. But yeah, maybe, it is better to not support `.ecrc.json`, as we're probably dropping support for it in the next major version.

README.md
theoludwig: Okay, yeah, good point.
>  trading readability for brevity. And this is exactly the pattern we're currently fixing here.

`.editorconfig-checker.json` should be good. :smile:

klaernie (changes requested):
The technical implementation of searching for config files is correct this way, I would only request that we search for our ideal config file name first (`.editorconfig-checker(rc)?.json`), then drop the `.json` suffix (if we actually want to encourage that behaviour) and that fail back to loading `.ecrc`.

This way users can test their migration to the new file format, and do not get surprised by adding a new config with a syntax error, checking their CI and noticing no problems, then drop their existing config and suddenly run into a problem as the syntax error now gets noticed.

Also having one filename would be better for adding a JSON schema (#194)

README.md
```diff
@@ -150,7 +150,10 @@ The following output formats are supported:
 
 ## Configuration
 
-The configuration is done via arguments or an `.ecrc` file.
+The configuration is done via arguments or it will take the first config file found with the following file names:
+
+- `.ecrc`
+- `.editorconfig-checker.json`
 
 A sample `.ecrc` file can look like this and will be used from your current working directory if not specified via the `--config` argument:
```
ccoVeille, line 158:
```suggestion
A sample configuration file can look like this and will be used from your current working directory if not specified via the `--config` argument:
```

cmd/editorconfig-checker/main.go
```diff
@@ -18,8 +18,8 @@ import (
 // version is dynamically set at compiletime
 var version string
 
-// defaultConfigFilePath determines where the config is located
-const defaultConfigFilePath = ".ecrc"
+// defaultConfigFileNames determines the file names where the config is located
+var defaultConfigFileNames = []string{".ecrc", ".editorconfig-checker.json"}
```
ccoVeille, line 22: Shouldn't it be the opposite?


```suggestion
var defaultConfigFileNames = []string{".editorconfig-checker.json", ".ecrc"}
```

README.md
klaernie: we should replace all mentions of the `.ecrc` with "configuration file", since when using `--config` users could name their file arbitrary names.

cmd/editorconfig-checker/main.go
theoludwig: Performance wise, it doesn't change much. And if we first check `.ecrc`, we optimize for existing users, if we first check `.editorconfig-checker.json`, we optimize for users migrating to the new config file name.

README.md
theoludwig: Updated. :smile:

cmd/editorconfig-checker/main.go
ccoVeille: Objection! 😅😁😂



The performance argument is not acceptable.

The file will be accessed once in a software that literally opens several, if not almost all,  files in a repository. And read all their content.

So no, a file not found hit is not a problem.

For me, the question is clearly functional.

Anyone starting using the new filename must have the settings contained in it, even there is .ecrc file in the folder

theoludwig: > The performance argument is not acceptable.
> 
> The file will be accessed once in a software that literally opens several, if not almost all, files in a repository. And read all their content.
> 
> So no, a file not found hit is not a problem.

Yes, I'm not against this change, but I don't understand why it matters, was thinking about "performance", but as we both agree, it pretty much doesn't change anything. 

We might have a deprecation notice for `.ecrc`. But we don't support merging of configuration files anyway, so the users should not use both `.ecrc` and `.editorconfig-checker.json`, so we might also warn the user, that only 1 config file is taken. Anyway, we will be able to simplify the code again, to only support `.editorconfig-checker.json` in the next major version, so it doesn't seems worth it to "work on it too much".

Anyway, I applied your suggestion.

README.md
```diff
@@ -169,9 +169,12 @@ The following output formats are supported:
 
 ## Configuration
 
-The configuration is done via arguments or an `.ecrc` file.
+The configuration is done via arguments or it will take the first config file found with the following file names:
 
-A sample `.ecrc` file can look like this and will be used from your current working directory if not specified via the `--config` argument:
+- `.ecrc`
+- `.editorconfig-checker.json`
```
ccoVeille, line 175:
```suggestion
- `.editorconfig-checker.json`
- `.ecrc` (deprecated filename, soon unsupported)
```

ccoVeille (approved):
except a side remark on the documentation, I'm fine with current implementation

klaernie (approved):
Except for the order in the README, good to go!

ccoVeille (approved):

PR #377, opened 2024-09-08
refactor: output FormatErrors

### LINKED ISSUE #374: Refactor error.go `FormatErrors`
This function slowly becomes unreadable, so refactoring it has become necessary, especially once #367 has been merged.

Blocked by: #367
Feel free to assign it to me

klaernie:
With this set of changes we should have a much more readable way of defining new output formats:
1. copy and paste a function in errors.go
2. add the format to `FormatErrors`' dispatching table
3. register the new format in output_formats.go's `ValidOutputFormats`
4. adjust README to document for the users
5. run the test suite allowing the snapshots to be updated

Adjusting the argument help text and adjusting the test suite of error.go is now automatic.

### DIFF
--- a/cmd/editorconfig-checker/main.go
+++ b/cmd/editorconfig-checker/main.go
@@ -10,6 +10,7 @@ import (
 	"github.com/editorconfig-checker/editorconfig-checker/v3/pkg/error"
 	"github.com/editorconfig-checker/editorconfig-checker/v3/pkg/files"
 	"github.com/editorconfig-checker/editorconfig-checker/v3/pkg/logger"
+	"github.com/editorconfig-checker/editorconfig-checker/v3/pkg/outputformat"
 	"github.com/editorconfig-checker/editorconfig-checker/v3/pkg/utils"
 	"github.com/editorconfig-checker/editorconfig-checker/v3/pkg/validation"
 )
@@ -39,8 +40,8 @@ func init() {
 	flag.BoolVar(&c.ShowVersion, "version", false, "print the version number")
 	flag.BoolVar(&c.Help, "help", false, "print the help")
 	flag.BoolVar(&c.Help, "h", false, "print the help")
-	flag.StringVar(&c.Format, "format", "default", "specify the output format: default, gcc, github-actions, codeclimate")
-	flag.StringVar(&c.Format, "f", "default", "specify the output format: default, gcc, github-actions, codeclimate")
+	flag.TextVar(&c.Format, "format", outputformat.Default, "specify the output format: "+outputformat.GetArgumentChoiceText())
+	flag.TextVar(&c.Format, "f", outputformat.Default, "specify the output format: "+outputformat.GetArgumentChoiceText())
 	flag.BoolVar(&c.Verbose, "verbose", false, "print debugging information")
 	flag.BoolVar(&c.Verbose, "v", false, "print debugging information")
 	flag.BoolVar(&c.Debug, "debug", false, "print debugging information")
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -10,6 +10,7 @@ import (
 	"github.com/editorconfig/editorconfig-core-go/v2"
 
 	"github.com/editorconfig-checker/editorconfig-checker/v3/pkg/logger"
+	"github.com/editorconfig-checker/editorconfig-checker/v3/pkg/outputformat"
 	"github.com/editorconfig-checker/editorconfig-checker/v3/pkg/utils"
 )
 
@@ -86,7 +87,7 @@ type Config struct {
 	// CONFIG FILE
 	Version             string
 	Verbose             bool
-	Format              string
+	Format              outputformat.OutputFormat
 	Debug               bool
 	IgnoreDefaults      bool
 	SpacesAftertabs     bool
@@ -186,7 +187,7 @@ func (c *Config) Merge(config Config) {
 		c.Verbose = config.Verbose
 	}
 
-	if config.Format != "" {
+	if config.Format.IsValid() {
 		c.Format = config.Format
 	}
 
--- a/pkg/error/error.go
+++ b/pkg/error/error.go
@@ -88,7 +88,7 @@ func ConsolidateErrors(errors []ValidationError, config config.Config) []Validat
 }
 
 func FormatErrorsAsHumanReadable(errors []ValidationErrors, config config.Config) []logger.LogMessage {
-	var formatted_errors []logger.LogMessage
+	var logMessages []logger.LogMessage
 
 	for _, fileErrors := range errors {
 		if len(fileErrors.Errors) == 0 {
@@ -97,33 +97,33 @@ func FormatErrorsAsHumanReadable(errors []ValidationErrors, config config.Config
 
 		relativeFilePath, err := files.GetRelativePath(fileErrors.FilePath)
 		if err != nil {
-			formatted_errors = append(formatted_errors, logger.LogMessage{Level: "error", Message: err.Error()})
+			logMessages = append(logMessages, logger.LogMessage{Level: "error", Message: err.Error()})
 			continue
 		}
 
 		fileErrors.Errors = ConsolidateErrors(fileErrors.Errors, config)
 
-		formatted_errors = append(formatted_errors, logger.LogMessage{Level: "warning", Message: fmt.Sprintf("%s:", relativeFilePath)})
+		logMessages = append(logMessages, logger.LogMessage{Level: "warning", Message: fmt.Sprintf("%s:", relativeFilePath)})
 		for _, singleError := range fileErrors.Errors {
 			if singleError.LineNumber == -1 {
-				formatted_errors = append(formatted_errors, logger.LogMessage{Level: "error", Message: fmt.Sprintf("\t%s", singleError.Message)})
+				logMessages = append(logMessages, logger.LogMessage{Level: "error", Message: fmt.Sprintf("\t%s", singleError.Message)})
 				continue
 			}
 
 			if singleError.AdditionalIdenticalErrorCount == 0 {
-				formatted_errors = append(formatted_errors, logger.LogMessage{Level: "error", Message: fmt.Sprintf("\t%d: %s", singleError.LineNumber, singleError.Message)})
+				logMessages = append(logMessages, logger.LogMessage{Level: "error", Message: fmt.Sprintf("\t%d: %s", singleError.LineNumber, singleError.Message)})
 				continue
 			}
 
-			formatted_errors = append(formatted_errors, logger.LogMessage{Level: "error", Message: fmt.Sprintf("\t%d-%d: %s", singleError.LineNumber, singleError.LineNumber+singleError.AdditionalIdenticalErrorCount, singleError.Message)})
+			logMessages = append(logMessages, logger.LogMessage{Level: "error", Message: fmt.Sprintf("\t%d-%d: %s", singleError.LineNumber, singleError.LineNumber+singleError.AdditionalIdenticalErrorCount, singleError.Message)})
 		}
 	}
 
-	return formatted_errors
+	return logMessages
 }
 
 func FormatErrorsAsGHA(errors []ValidationErrors, config config.Config) []logger.LogMessage {
-	var formatted_errors []logger.LogMessage
+	var logMessages []logger.LogMessage
 
 	for _, fileErrors := range errors {
 		if len(fileErrors.Errors) == 0 {
@@ -132,7 +132,7 @@ func FormatErrorsAsGHA(errors []ValidationErrors, config config.Config) []logger
 
 		relativeFilePath, err := files.GetRelativePath(fileErrors.FilePath)
 		if err != nil {
-			formatted_errors = append(formatted_errors, logger.LogMessage{Level: "error", Message: err.Error()})
+			logMessages = append(logMessages, logger.LogMessage{Level: "error", Message: err.Error()})
 			continue
 		}
 
@@ -141,25 +141,25 @@ func FormatErrorsAsGHA(errors []ValidationErrors, config config.Config) []logger
 		// github-actions: A format dedicated for usage in Github Actions
 		for _, singleError := range fileErrors.Errors {
 			if singleError.LineNumber == -1 {
-				formatted_errors = append(formatted_errors, logger.LogMessage{Level: "error", Message: fmt.Sprintf("::error file=%s::%s", relativeFilePath, singleError.Message)})
+				logMessages = append(logMessages, logger.LogMessage{Level: "error", Message: fmt.Sprintf("::error file=%s::%s", relativeFilePath, singleError.Message)})
 				continue
 			}
 
 			if singleError.AdditionalIdenticalErrorCount == 0 {
-				formatted_errors = append(formatted_errors, logger.LogMessage{Level: "error", Message: fmt.Sprintf("::error file=%s,line=%d::%s", relativeFilePath, singleError.LineNumber, singleError.Message)})
+				logMessages = append(logMessages, logger.LogMessage{Level: "error", Message: fmt.Sprintf("::error file=%s,line=%d::%s", relativeFilePath, singleError.LineNumber, singleError.Message)})
 				continue
 			}
 
-			formatted_errors = append(formatted_errors, logger.LogMessage{Level: "error", Message: fmt.Sprintf("::error file=%s,line=%d,endLine=%d::%s", relativeFilePath, singleError.LineNumber, singleError.LineNumber+singleError.AdditionalIdenticalErrorCount, singleError.Message)})
+			logMessages = append(logMessages, logger.LogMessage{Level: "error", Message: fmt.Sprintf("::error file=%s,line=%d,endLine=%d::%s", relativeFilePath, singleError.LineNumber, singleError.LineNumber+singleError.AdditionalIdenticalErrorCount, singleError.Message)})
 		}
 	}
 
-	return formatted_errors
+	return logMessages
 }
 
 // gcc: A format mimicking the error format from GCC.
 func FormatErrorsAsGCC(errors []ValidationErrors, config config.Config) []logger.LogMessage {
-	var formatted_errors []logger.LogMessage
+	var logMessages []logger.LogMessage
 
 	for _, fileErrors := range errors {
 		if len(fileErrors.Errors) == 0 {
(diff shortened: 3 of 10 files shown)


cmd/editorconfig-checker/main.go
```diff
@@ -10,6 +10,7 @@ import (
 	"github.com/editorconfig-checker/editorconfig-checker/v3/pkg/error"
 	"github.com/editorconfig-checker/editorconfig-checker/v3/pkg/files"
 	"github.com/editorconfig-checker/editorconfig-checker/v3/pkg/logger"
+	"github.com/editorconfig-checker/editorconfig-checker/v3/pkg/output_formats"
```
ccoVeille, line 13: snake_case in a package name :grimacing: 

https://go.dev/blog/package-names

> Good package names are short and clear. They are lower case, with no under_scores or mixedCaps. They are often simple nouns, such as:

pkg/config/config.go
```diff
@@ -86,7 +87,7 @@ type Config struct {
 	// CONFIG FILE
 	Version             string
 	Verbose             bool
-	Format              string
+	Format              output_formats.OutputFormat
```
ccoVeille, line 90: if you have output_format but then call a variable OutputFormat there is something wrong

pkg/error/error.go
```diff
@@ -86,72 +87,114 @@ func ConsolidateErrors(errors []ValidationError, config config.Config) []Validat
...
 }
 
-// FormatErrors prints the errors to the console
-func FormatErrors(errors []ValidationErrors, config config.Config) []logger.LogMessage {
+func FormatErrorsAsHumanReadable(errors []ValidationErrors, config config.Config) []logger.LogMessage {
 	var formatted_errors []logger.LogMessage
-	var codeclimateIssues []CodeclimateIssue
 
 	for _, fileErrors := range errors {
 		if len(fileErrors.Errors) > 0 {
```
ccoVeille, line 94: OH MY GOD the complexity of this loop

I know it's not your fault, but could you at least do this

```suggestion
		if len(fileErrors.Errors) == 0 {
			continue
		}
```

It would decrease the number of level

pkg/error/error.go
```diff
@@ -86,72 +87,114 @@ func ConsolidateErrors(errors []ValidationError, config config.Config) []Validat
...
+	}
+
+	return formatted_errors
+}
+
+func FormatErrorsAsGHA(errors []ValidationErrors, config config.Config) []logger.LogMessage {
+	var formatted_errors []logger.LogMessage
+
+	for _, fileErrors := range errors {
+		if len(fileErrors.Errors) > 0 {
```
ccoVeille, line 125: same

```suggestion
		if len(fileErrors.Errors) == 0 {
			continue
		}
```

pkg/error/error.go
```diff
@@ -86,72 +87,114 @@ func ConsolidateErrors(errors []ValidationError, config config.Config) []Validat
...
 
+	return formatted_errors
+}
+
+// gcc: A format mimicking the error format from GCC.
+func FormatErrorsAsGCC(errors []ValidationErrors, config config.Config) []logger.LogMessage {
+	var formatted_errors []logger.LogMessage
+
+	for _, fileErrors := range errors {
+		if len(fileErrors.Errors) > 0 {
```
ccoVeille, line 157: same

```suggestion
		if len(fileErrors.Errors) == 0 {
			continue
		}
```

pkg/error/error.go
```diff
@@ -86,72 +87,114 @@ func ConsolidateErrors(errors []ValidationError, config config.Config) []Validat
...
 			}
 		}
 	}
 
+	return formatted_errors
+}
+
+// gcc: A format mimicking the error format from GCC.
+func FormatErrorsAsGCC(errors []ValidationErrors, config config.Config) []logger.LogMessage {
+	var formatted_errors []logger.LogMessage
```
ccoVeille, line 154: there is variable with underscore in Go

pkg/error/error.go
```diff
@@ -86,72 +87,114 @@ func ConsolidateErrors(errors []ValidationError, config config.Config) []Validat
...
+	for _, fileErrors := range errors {
+		if len(fileErrors.Errors) > 0 {
+			relativeFilePath, err := files.GetRelativePath(fileErrors.FilePath)
+			if err != nil {
+				formatted_errors = append(formatted_errors, logger.LogMessage{Level: "error", Message: err.Error()})
+				continue
+			}
+
+			for _, singleError := range fileErrors.Errors {
+				var lineNo = 0
```
ccoVeille, line 165: Idiomatic way would be either

```suggestion
				lineNo := 0
```
or

```suggestion
				var lineNo int
```

pkg/error/error.go
```diff
@@ -86,72 +87,114 @@ func ConsolidateErrors(errors []ValidationError, config config.Config) []Validat
...
+	return formatted_errors
+}
+
+// codeclimate: A format that is compatible with the codeclimate format for GitLab CI.
+// https://docs.gitlab.com/ee/ci/testing/code_quality.html#implement-a-custom-tool
+func FormatErrorsAsCodeclimate(errors []ValidationErrors, config config.Config) []logger.LogMessage {
+	var codeclimateIssues []CodeclimateIssue
+
+	for _, fileErrors := range errors {
+		if len(fileErrors.Errors) > 0 {
```
ccoVeille, line 183: same

```suggestion
		if len(fileErrors.Errors) == 0 {
			continue
		}
```

pkg/error/error.go
```diff
@@ -86,72 +87,114 @@ func ConsolidateErrors(errors []ValidationError, config config.Config) []Validat
...
+				if singleError.LineNumber != -1 {
+					if singleError.AdditionalIdenticalErrorCount != 0 {
+						formatted_errors = append(formatted_errors, logger.LogMessage{Level: "error", Message: fmt.Sprintf("\t%d-%d: %s", singleError.LineNumber, singleError.LineNumber+singleError.AdditionalIdenticalErrorCount, singleError.Message)})
 					} else {
-						formatted_errors = append(formatted_errors, logger.LogMessage{Level: "error", Message: fmt.Sprintf("::error file=%s::%s", relativeFilePath, singleError.Message)})
+						formatted_errors = append(formatted_errors, logger.LogMessage{Level: "error", Message: fmt.Sprintf("\t%d: %s", singleError.LineNumber, singleError.Message)})
 					}
+				} else {
+					formatted_errors = append(formatted_errors, logger.LogMessage{Level: "error", Message: fmt.Sprintf("\t%s", singleError.Message)})
 				}
```
ccoVeille, line 113: this piece of code is too complex, it can be simplified

first, no matter the branch, you will end in appending to format_errors

So let try this please

```go
			for _, singleError := range fileErrors.Errors {
				logMessage := fmt.Sprintf("\t%s", singleError.Message)
				if singleError.LineNumber != -1 {
					logMessage = fmt.Sprintf("\t%d: %s", singleError.LineNumber, singleError.Message)}
					if singleError.AdditionalIdenticalErrorCount != 0 {
						logMessage = fmt.Sprintf("\t%d-%d: %s", singleError.LineNumber, singleError.LineNumber+singleError.AdditionalIdenticalErrorCount, singleError.Message)
					}
				}
				formatted_errors = append(formatted_errors, logger.LogMessage{Level: "error", Message: logMessage)
			}
```

This pattern could be applied everywhere it would increase the readability and maintenability

pkg/error/error.go
```diff
@@ -86,72 +87,114 @@ func ConsolidateErrors(errors []ValidationError, config config.Config) []Validat
...
+					if singleError.AdditionalIdenticalErrorCount != 0 {
+						formatted_errors = append(formatted_errors, logger.LogMessage{Level: "error", Message: fmt.Sprintf("::error file=%s,line=%d,endLine=%d::%s", relativeFilePath, singleError.LineNumber, singleError.LineNumber+singleError.AdditionalIdenticalErrorCount, singleError.Message)})
 					} else {
-						formatted_errors = append(formatted_errors, logger.LogMessage{Level: "error", Message: fmt.Sprintf("\t%s", singleError.Message)})
+						formatted_errors = append(formatted_errors, logger.LogMessage{Level: "error", Message: fmt.Sprintf("::error file=%s,line=%d::%s", relativeFilePath, singleError.LineNumber, singleError.Message)})
 					}
+				} else {
+					formatted_errors = append(formatted_errors, logger.LogMessage{Level: "error", Message: fmt.Sprintf("::error file=%s::%s", relativeFilePath, singleError.Message)})
 				}
 			}
```
ccoVeille, line 145: same logic here

```go
			for _, singleError := range fileErrors.Errors {
				logMessage := fmt.Sprintf("::error file=%s::%s", relativeFilePath, singleError.Message)
				if singleError.LineNumber != -1 {
					logMessage = fmt.Sprintf("::error file=%s,line=%d::%s", relativeFilePath, singleError.LineNumber, singleError.Message)
					if singleError.AdditionalIdenticalErrorCount != 0 {
						logMessage = fmt.Sprintf("::error file=%s,line=%d,endLine=%d::%s", relativeFilePath, singleError.LineNumber, singleError.LineNumber+singleError.AdditionalIdenticalErrorCount, singleError.Message)
				}
				formatted_errors = append(formatted_errors, logger.LogMessage{Level: "error", Message: logMessage})
			}
```

pkg/output_formats/output_formats.go
```diff
@@ -0,0 +1,55 @@
...
+	var output_strings []string
+	for _, f := range ValidOutputFormats {
+		output_strings = append(output_strings, string(f))
+	}
+	return strings.Join(output_strings, ", ")
+}
+
+func (format OutputFormat) MarshalText() ([]byte, error) {
+	if !format.IsValid() {
+		return nil, errors.New(string(format) + " is not a valid output format")
```
ccoVeille, line 35:
```suggestion
		return nil, fmt.Errorf("%s is not a valid output format", tring(format))
```

if not

```suggestion
		return nil, fmt.Errorf("%q is not a valid output format", tring(format))
```

pkg/output_formats/output_formats.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+	return []byte(format), nil
+}
+
+func (format *OutputFormat) UnmarshalText(data []byte) error {
+	*format = OutputFormat(string(data))
+	if !format.IsValid() {
+		return errors.New(string(data) + " is not a valid output format")
```
ccoVeille, line 43: same, you should use fmt.Errorf

pkg/output_formats/output_formats.go
```diff
@@ -0,0 +1,55 @@
...
+	return nil
+}
+
+func (format OutputFormat) IsValid() bool {
+	for _, supportedFormat := range ValidOutputFormats {
+		if format == supportedFormat {
+			return true
+		}
+	}
+	return false
```
ccoVeille, line 54: this is [slices.Contains](https://pkg.go.dev/slices#Contains)

ccoVeille (changes requested):
thanks for the changes

here is my review

pkg/output_formats/output_formats.go
klaernie: How TF did miss that while googling explicitly for it... 🤦🏻

ccoVeille: slices package is somehow one year old, Internet is full of example on how to code it (but posted before slices.Contains existed)

cmd/editorconfig-checker/main.go
klaernie: Okay, would `outputformats` be a suitable name?

pkg/error/error.go
ccoVeille: It would be a good idea, yes. 

But I would discourage you to do it in this PR

one step at a time :tm: 

So another issue, and another PR for later

pkg/config/config.go
klaernie: Is there a name you would prefer for this type? I cannot think of one better representing a type of output supported by editorconfig-checker.

klaernie:
I'll tackle the technical debt tomorrow

cmd/editorconfig-checker/main.go
ccoVeille: `format` would be simpler and more classic

but `outputformat` is OK (but singular)

Yes, there is already package with a plural, but they are not respecting the idiomatic way to name them

> Package names are usually singular except the cases when they collide with reserved words like strings, types, etc.

ccoVeille:
> I'll tackle the technical debt tomorrow

Good luck, have fun

cmd/editorconfig-checker/main.go
klaernie: fixed

pkg/error/error.go
klaernie: done

pkg/error/error.go
klaernie: done

pkg/error/error.go
klaernie: done

pkg/error/error.go
klaernie: I pulled a lot of complexity out of these ones by switching to an early exit style - wish I had thought of that when I introduced half of that complexity a couple weeks back, but no point crying about spilled tea.. ;) Thankfully flipping these conditions was easy when the IDE provides a way to do it automatically.

pkg/output_formats/output_formats.go
klaernie: done

pkg/output_formats/output_formats.go
klaernie: done

pkg/output_formats/output_formats.go
klaernie: hm.. I should dig up that stackoverflow I read and fix it.. but anyway, I replaced it

pkg/error/error.go
klaernie: done

pkg/error/error.go
klaernie: done

pkg/error/error.go
klaernie: found a name that is way better..

pkg/error/error.go
klaernie: done

pkg/config/config.go
klaernie: don't really have an idea if you would like to still see a change here, it being `outputformat.OutputFormat` now.

klaernie:
> > I'll tackle the technical debt tomorrow
> 
> Good luck, have fun

Actually, it was not that painful - almost enjoyable compared to coming up with a way to code enums. But a lot of that is having a testsuite that covers a lot - quite the difference to at work. Thanks for the many pointers!

Before merging I'll definitely rebase again and try to reduce the number of commits to a sensible amount, but for the purpose of reviewing I left them as they are right now.

pkg/config/config.go
ccoVeille: I don't have either

pkg/error/error.go
```diff
@@ -86,69 +87,123 @@ func ConsolidateErrors(errors []ValidationError, config config.Config) []Validat
...
+
+			logmsgs = append(logmsgs, logger.LogMessage{Level: "error", Message: fmt.Sprintf("\t%d-%d: %s", singleError.LineNumber, singleError.LineNumber+singleError.AdditionalIdenticalErrorCount, singleError.Message)})
+		}
+	}
+
+	return logmsgs
+}
+
+func FormatErrorsAsGHA(errors []ValidationErrors, config config.Config) []logger.LogMessage {
+	var logmsgs []logger.LogMessage
```
ccoVeille, line 126:
```suggestion
	var logMessages []logger.LogMessage
```

or


```suggestion
	var logMsgs []logger.LogMessage
```

I prefer the first one

pkg/error/error.go
```diff
@@ -86,69 +87,123 @@ func ConsolidateErrors(errors []ValidationError, config config.Config) []Validat
...
+			logmsgs = append(logmsgs, logger.LogMessage{Level: "error", Message: fmt.Sprintf("::error file=%s,line=%d,endLine=%d::%s", relativeFilePath, singleError.LineNumber, singleError.LineNumber+singleError.AdditionalIdenticalErrorCount, singleError.Message)})
+		}
+	}
+
+	return logmsgs
+}
+
+// gcc: A format mimicking the error format from GCC.
+func FormatErrorsAsGCC(errors []ValidationErrors, config config.Config) []logger.LogMessage {
+	var logmsgs []logger.LogMessage
```
ccoVeille, line 162: Same

pkg/error/error_test.go
```diff
@@ -315,18 +315,11 @@ func TestFormatErrors(t *testing.T) {
...
-	config2 := config.Config{Format: "gcc"}
-	s.MatchSnapshot(t, FormatErrors(input, config2))
-
-	config3 := config.Config{Format: "github-actions"}
-	s.MatchSnapshot(t, FormatErrors(input, config3))
-
-	config4 := config.Config{Format: "codeclimate"}
-	s.MatchSnapshot(t, FormatErrors(input, config4))
+	for _, format := range outputformat.ValidOutputFormats {
+		t.Run(string(format), func(t *testing.T) {
```
ccoVeille, line 319: Add a fmt.Stringer interface to your OutputFormat type

This way you could use format without casting the type here

It's as simple as this

```go
func (f OutputFormat) String() string {
    return string(f)
}
```

pkg/outputformat/outputformat.go
```diff
@@ -0,0 +1,51 @@
+// Package for having structured access to our output formats
+package outputformat
+
+import (
+	"fmt"
+	"slices"
+	"strings"
+)
+
+type OutputFormat string
```
ccoVeille, line 10: Add a stringer to this type, see https://github.com/editorconfig-checker/editorconfig-checker/pull/377#discussion_r1751019023

ccoVeille (approved):

pkg/outputformat/outputformat.go
klaernie: done

pkg/error/error_test.go
klaernie: I added the interface, but `t.Run` insists on getting a string. It helped in two other spots in but sadly here it doesn't work:
> editorconfig-checker/pkg/error/error_test.go:319:9 cannot use format (variable of type outputformat.OutputFormat) as string value in argument to t.Run (compile)

pkg/error/error.go
klaernie: done

pkg/error/error.go
klaernie: done

theoludwig (approved):
LGTM, thanks! :smile:

pkg/error/error_test.go
ccoVeille: Ob that's right!

ccoVeille (approved):

klaernie:
I rebased the commits into a set I can live with, so now I'd suggest we let these ripe for a few days for anyone to review before merging.

Follow up changeset will be reworking the interface currently done using the `LogMessage`, since passing an array of `LogMessage` back to main is a bit pointless, only for them to be immediately output. The replacement will be an io.Writer stored in `Logger`, initialized to os.Stdout.

PR #378, opened 2024-09-09
docs: add CONTRIBUTING.md and MAINTAINERS.md

theoludwig:
As the project is gaining activity again, and new contributors arise, it might be the right moment to define what are the "rules" for contributing to this project, and how pull requests are reviewed and merged.

Ref: #364 

(please take note, that English is not my primary language, so feel free to correct me :smile:)

Issue #82, opened 2019-08-18
Add autocompletion for shells


mstruebing:

ccoVeille:
any objection to move away from std lib flag and move to [cobra](https://github.com/spf13/cobra) ?

https://github.com/editorconfig-checker/editorconfig-checker/blob/main/cmd/editorconfig-checker/main.go

It should be "easy" :tm: (famous last words)

What do you guy think ?

I'm not asking for people to code it, but for an agreement about the idea to use [cobra](https://github.com/spf13/cobra)

klaernie:
I was considering the same - but one thing that kept me from suggesting it was my perception of it requiring some restructuring - but I might be mistaken there. In my messy #380 branch I needed to restructure parsing already to make it possible to call main multiple times from the test cases.

theoludwig:
If possible, before doing any major refactor, we should add comprehensive e2e tests which are not dependent of the framework used behind the scene, so we can make these major refactor "without fear", or at least with more confidence.
Even better, we could make the tests agnostic of the programming languages with [bats](https://bats-core.readthedocs.io/en/stable/index.html).

I don't know what it takes to replace std lib by [cobra](https://github.com/spf13/cobra), and if it allows to support more easily autocompletion for shells, but it seems like major refactoring work with code which is not necessarily well tested, or with test code too much coupled with implementation details.

ccoVeille:
You raised interesting points.

While they are valid, it seems to me it means this issue should be updated, because it doesn't sound like a "good first issue"

klaernie:
I think these are two quite different aspects. IMHO with #380 we gain a lot of testing of our main function - about 95% according to my last run of `make test`.

Introducing a non-golang testing framework makes contributing rather harder that easier, and I think no matter the commandline argument parsing library used we can test-drive most of our functionality by providing directories filled with test data and a matching config file and feeding it to `main()` with `os.Args`. Capturing the exact output (of `main`) is something not done yet (although it is done for testing all out output formats in the `error` package), but we are not exactly missing much there in terms of output lines not verified against a snapshot.

Using an external framework could prove useful to only run it in the CI to independently verify that the most common case still works and was not broken by a test case - but our coverage checks already complain loudly when test coverage reduces.

klaernie:
Oh, and just to be clear: my work in #380 only needed to separate the flag definition from the actual parsing, as one may not define the flag set more than once, but we did 50% of parsing arguments in the `init` function instead of in `main`.

PR #380, opened 2024-09-15
refactor: replace Logger.LogMessage by storing an io.Writer

klaernie:
This replaces the bad design I made in #360 regarding the output of formerly error.PrintErrors, which became error.FormatErrors and is now again error.PrintErrors.

Passing around slices of `LogMessage` what not the best design - it allowed testing the new code, but the resulting code was unreadable.

Now we have a `Logger.writer` that automatically gets initialized to `os.Stdout`, but can be changed using `Logger.SetWriter()`. This allows us to restore the previous form of calling `Logger.Error` and friends directly, and still capture their output to make the tests work.

Along the way I noticed a few inconsistencies that needed to be fixed, I hope this PR is still reviewable.


pkg/logger/logger.go
```diff
@@ -13,115 +15,103 @@ const (
...
 
-type LogMessage struct {
-	Level   string
-	Message string
-}
-
 // Logger struct
 type Logger struct {
 	Verbosee bool
 	Debugg   bool
```
ccoVeille, line 21: I know it's out of the current PR scope

But please consider adding a comment if it's "normal" to have Verbosee and Debugg

I'm sure it's possible to find a better suffix. Here it looks like a typo

pkg/logger/logger.go
```diff
@@ -13,115 +15,103 @@ const (
...
-	fmt.Printf("%s", message)
-}
-
-// Println prints a message with a trailing newline
-func Println(message string) {
-	fmt.Printf("%s\n", message)
+// println prints a message with a trailing newline
+func (l Logger) println(message string) {
+	l.LazyInit()
+	fmt.Fprintf(l.writer, "%s\n", message)
```
ccoVeille, line 104:
```suggestion
	fmt.Fprintln(l.writer, message)
```

pkg/logger/logger.go
```diff
@@ -13,115 +15,103 @@ const (
...
 }
 
-// PrintColor prints a message in a given ANSI-color
-func PrintColor(message string, color string) {
-	fmt.Printf("%s%s%s", color, message, RESET)
+// printColor prints a message in a given ANSI-color
+func (l Logger) printColor(message string, color string) {
+	l.LazyInit()
+	fmt.Fprintf(l.writer, "%s%s%s", color, message, RESET)
```
ccoVeille, line 110: RESET variable looks abnormal (its case)

First a full uppercase is wrongly discouraged in Go

Then it also means it's exported.

Do we really need this?

But this remark is out of PR scope as RED, YELLOW, RESET … were present before this current PR

pkg/logger/logger.go
```diff
@@ -13,115 +15,103 @@ const (
...
+	return logger
+}
+
+// initialize the Logger to write to standard output
+func (l *Logger) Init() {
+	l.writer = os.Stdout
+}
+
+// ensure the Logger is initialized on first print
+func (l *Logger) LazyInit() {
```
ccoVeille, line 38:
```suggestion
func (l *Logger) lazyInit() {
```

I don't think you need it to be exported

pkg/logger/logger.go
klaernie: I renamed them to `VerboseEnabled` and `DebugEnabled` - that seemed like a better name and avoiding that typo-esque suffix

pkg/logger/logger.go
klaernie: good point. done.

pkg/logger/logger.go
klaernie: I renamed them by normalizing their names and prefixing them with `escSeq` which should help people unfamiliar with escape sequences to recognize them as such.

pkg/logger/logger.go
klaernie: You're totally right. fixed.

Makefile
```diff
@@ -42,7 +42,7 @@ uninstall: ## Remove executable from PATH
 	rm -f $(DESTDIR)$(prefix)$(mandir)/man1/editorconfig-checker.1
 
 test: ## Run test suite
-	@go test -race -coverprofile=coverage.txt -covermode=atomic ./...
+	@go test -ldflags "-X github.com/editorconfig-checker/editorconfig-checker/v3/cmd/editorconfig-checker.version=$(CURRENT_VERSION)" -race -coverprofile=coverage.txt -covermode=atomic ./...
```
theoludwig, line 45: Could you explain why this change is needed?

cmd/editorconfig-checker/main.go
```diff
@@ -2,56 +2,69 @@
...
 	"github.com/editorconfig-checker/editorconfig-checker/v3/pkg/config"
 	"github.com/editorconfig-checker/editorconfig-checker/v3/pkg/error"
 	"github.com/editorconfig-checker/editorconfig-checker/v3/pkg/files"
-	"github.com/editorconfig-checker/editorconfig-checker/v3/pkg/logger"
 	"github.com/editorconfig-checker/editorconfig-checker/v3/pkg/outputformat"
 	"github.com/editorconfig-checker/editorconfig-checker/v3/pkg/utils"
 	"github.com/editorconfig-checker/editorconfig-checker/v3/pkg/validation"
 )
 
 // version is used vor the help
```
theoludwig, line 19: Not related to this PR, but now that I see the typo, and that the PR is related to refactoring and code clean up anyway, might be worth to fix it:

```suggestion
// version is used for the help
```

cmd/editorconfig-checker/main_test.go
theoludwig, line 1: Good to add tests for the main, and CLI options. As explained in https://github.com/editorconfig-checker/editorconfig-checker/issues/82#issuecomment-2361817709, it would be even better to test it agnostic to the framework or even more agnostic, to the programming languages used.

cmd/editorconfig-checker/main.go
klaernie: lol.. I know - once one sees it, one cannot unsee it.. I fixed it, and added a little to the explanation on what that variable is for.

Makefile
klaernie: This is one of the changes I had to make due to the issue I described in #383 - since with that PR we are starting to run the binary (or at least it's `main` function) we also tend to parse config files - which means we need to have the current version available for comparison - hence the injection of the overwriting the initializing value of that variable.

cmd/editorconfig-checker/main_test.go
klaernie: I think this currently is as far as possible testing from outside the main function - yes, we hook into the `main` function by overriding it's `exitStub`, but in every other perspective we run `main` as a goroutine as independent of the test framework as possible - providing only `os.Args` and external files - calling main and collecting only the return code. I did not yet attempt to override `os.Stdout` for capturing end-to-end snapshot tests (that might become its own PR later on).

Makefile
theoludwig: Alright, thank you for the explanation. :+1:

cmd/editorconfig-checker/main_test.go
theoludwig: Yes, let's keep the tests in Go for the moment, easier to track test coverage this way also.

theoludwig (approved):

klaernie:
I should probably rebase this branch, since there were some updates to main. But I wonder - is this PR to big, and I should try to break it into a few PRs? Sadly I think I coded myself into a corner with that big of a change set..

theoludwig:
> I should probably rebase this branch, since there were some updates to main. But I wonder - is this PR to big, and I should try to break it into a few PRs? Sadly I think I coded myself into a corner with that big of a change set..

Yeah, the PR is quite big, and quite hard to review with all the changes. But I think, it's fine now, I think before doing anything, we need to wait other reviewers comments.
Or as Max said, 2 reviewers needed for a PR might be "too much" (https://github.com/editorconfig-checker/editorconfig-checker/pull/378#pullrequestreview-2297480255), we might reduce it to 1.
This PR is already opened since 1 month, we might still wait 1 week, and if next week, there are still no activity from others people, I will open the PR for reducing the amount of reviewer needed for a PR, and then we can merge this one.
Do you think it's a good way to move forward?

Also, do you think, we should wait #385 to release v3.1.0? I feel like there are already lot of changes waiting for a little while to be released, and we might release the recently added features, rather sooner than later.
Is #385 "ready" for most of the work? If yes, why still in draft? You might undraft it, so we can start reviewing more in depth, and working on it.
(I know it's off topic to talk about release process in this PR, but would like to talk about it here, as it's easier)

You can rebase if you want, but there are no conflicts and the difference with `main` is very little, it should have no impacts  with this PR.

klaernie:
> Also, do you think, we should wait #385 to release v3.1.0? I feel like there are already lot of changes waiting for a little while to be released, and we might release the recently added features, rather sooner than later. Is #385 "ready" for most of the work? If yes, why still in draft? You might undraft it, so we can start reviewing more in depth, and working on it. (I know it's off topic to talk about release process in this PR, but would like to talk about it here, as it's easier)

I think it would be easiest to stuff it all into a v3.2.0 - #385 and #371 kind of go hand in hand - #371 takes care of the Makefile-based builds, and #385 revamps the official release process to entire be driven by goreleaser, making loads of the Makefile irrelevant. But for both we first need to ensure the wrapper projects are aware of the name change. On the other hand having an automatic release process would be nice to iterate quicker. I actually forgot to set it out of Draft status, my "check the existing archives" was my target for it being reviewable, but I totally overlooked that.
(I'm gonna copy that over to #385)

Your way forward with this branch sounds about right - I think the biggest problem is that even a small team not always has three people available (one to code, two to review), so it can get lengthy sometimes. And this only quadruples, when it's 32 commits and 1000 line changes, simply increasing the mental "do I really want to tackle this big PR now?" ;) Well, I definitely learned from that again.

cmd/editorconfig-checker/main_test.go
```diff
@@ -0,0 +1,130 @@
...
+
+func captureReturncode(code int) {
+	mainHasRun <- code
+	runtime.Goexit()
+}
+
+func setArguments(args ...string) {
+	// os.Args needs to have the binary
+	os.Args = append([]string{"editorconfig-checker"}, args...)
+}
```
ccoVeille, line 19: This is not a good way to do.

I would suggest this


```suggestion
func setArguments(t *testing.T, args ...string) {
	t.Helper()

	initialArgs := os.Args

	// os.Args needs to have the binary
	os.Args = append([]string{"editorconfig-checker"}, args...)
	t.Cleanup(func() {
		os.Args = initialArgs
	})
}
```

cmd/editorconfig-checker/main.go
```diff
@@ -2,56 +2,69 @@
...
 var currentConfig *config.Config
 
-// Init function, runs on start automagically
-func init() {
-	var configFilePath string
-	var tmpExclude string
-	var c config.Config
-	var init bool
+// exitStub is there to be replaced while running the tests
+var exitStub = os.Exit
```
ccoVeille, line 30: I won't call it stub here. What you do in the tests is a replacing it with a stub.

The calls made in main func with "exitStub" are very strange to me

Here it's the name that will be called in the code.

I could puld use something like this


```suggestion
var exiter = os.Exit
```

```suggestion
var exitFunc = os.Exit
```

theoludwig:
> You can rebase if you want, but there are no conflicts and the difference with main is very little, it should have no impacts with this PR.

This PR will now need to be rebased. I advise to wait #389 to be merged, so that you only resolve conflicts once.

cmd/editorconfig-checker/main_test.go
klaernie: Would you just call it bad style to not restore the original arguments, or do you have any practical concerns? The only other consumer of `os.Args` is the `testing` package itself, but it has completed it's `os.Args` parsing by then.

cmd/editorconfig-checker/main.go
klaernie: You're right, technically speaking this would make this function in main.go `exitProxy`.
I'm not really a fan of `exiter` since it's too close to the stem "exciting", and just appending `Func` is also not my cup of tea. I'd hope you can accept `exitProxy`.

cmd/editorconfig-checker/main_test.go
klaernie: I reworked it according to your suggestions

klaernie:
I rebased on top of the current main branch, incorporating the changes requested so far.

pkg/config/config.go
```diff
@@ -116,11 +116,14 @@ type DisabledChecks struct {
...
 func NewConfig(configPaths []string) (*Config, error) {
 	if len(configPaths) == 0 {
-		return nil, fmt.Errorf("No config paths provided")
+		return nil, fmt.Errorf("no config paths provided")
 	}
 
 	var config Config
 
+	// initialize the Logger early, so usage of it does not fail
+	config.Logger.Init()
```
ccoVeille, line 125: I would have added a Init to config, that would have called the Logger Init

This way, if the logger changes nothing needs to be updated outside Config

But why not, it's the NewConfig constructor

pkg/config/config.go
```diff
@@ -278,7 +284,7 @@ func (c Config) GetExcludesAsRegularExpression() string {
 // Save saves the config to it's Path
 func (c Config) Save(version string) error {
 	if utils.IsRegularFile(c.Path) {
-		return fmt.Errorf("File `%v` already exists", c.Path)
+		return fmt.Errorf("file `%v` already exists", c.Path)
```
ccoVeille, line 287: There is no reason to use %v for a string

So, this would be preferred

```suggestion
		return fmt.Errorf("file `%s` already exists", c.Path)
```

Also, this could be considered


```suggestion
		return fmt.Errorf("file %q already exists", c.Path)
```

cmd/editorconfig-checker/main.go
```diff
@@ -66,88 +79,90 @@ func init() {
 	currentConfig, _ = config.NewConfig(configPaths)
```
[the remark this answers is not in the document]
klaernie: it's not needed for `.NewConfig`, but for `.ParseConfig` - where it is present

cmd/editorconfig-checker/main.go
```diff
@@ -2,57 +2,70 @@
...
 	"github.com/editorconfig-checker/editorconfig-checker/v3/pkg/outputformat"
 	"github.com/editorconfig-checker/editorconfig-checker/v3/pkg/utils"
 	"github.com/editorconfig-checker/editorconfig-checker/v3/pkg/validation"
 )
 
-// version is used vor the help
+// version is used for the help and to verify against the version stored in the config file
 // version is dynamically set at compiletime
-var version string
+var version string = "v3.0.3"
```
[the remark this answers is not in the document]
klaernie: I would leave coming up with a better strategy up to #385 - here I just rebased my change to work nicely with #385 ( I introduced a magic string "development" as the version content and checked for that in a piece later, but this will no longer be needed after #385, maybe in two days time)

pkg/config/config.go
klaernie: I later switched to a lazy initialization, which is more reliable, so this was not needed at all. I removed it.

cmd/editorconfig-checker/main_test.go
```diff
@@ -0,0 +1,137 @@
...
+	lastSeenCode := <-mainHasRun
+	if lastSeenCode != 0 {
+		t.Errorf("main exited with return code %d, but we expected 0", lastSeenCode)
+	}
+}
+
+func TestMainInitializingANewConfig(t *testing.T) {
+	dir := t.TempDir()
+	setArguments(t, "--debug", "--verbose", "--init", "--config", dir+"/testwriteconfig.json")
+	go main()
```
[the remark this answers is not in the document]
klaernie: good point, done.

pkg/config/config.go
klaernie: I decided to keep the quotes - if for some obscure reason `c.Path` is empty or only filled with unprintable characters one would no longer recognize that this error message was supposed to contain a path.
But I replaced the default rendering with string rendering.

cmd/editorconfig-checker/main_test.go
```diff
@@ -0,0 +1,137 @@
...
+		os.Args = initialArgs
+	})
+
+	// os.Args needs to have the binary as the first element
+	os.Args = append([]string{"editorconfig-checker"}, args...)
+}
+
+func TestMainOurCodebase(t *testing.T) {
+	cdRelativeToRepo(t, "")
+	setArguments(t, "--debug", "--verbose", "--exclude", "\\.git", "--exclude", "\\.exe$")
```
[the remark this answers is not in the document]
klaernie: Done in all four strings.

cmd/editorconfig-checker/main.go
```diff
@@ -66,88 +79,90 @@ func init() {
...
 	config.Logger.Verbose("%d files checked", len(filePaths))
 
-	if errorCount != 0 {
-		os.Exit(1)
+	if error.GetErrorCount(errors) != 0 {
+		exitProxy(1)
 	}
 
-	os.Exit(0)
+	exitProxy(0)
```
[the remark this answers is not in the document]
klaernie: Done - including a test case that ensures nobody ever touches the values of these constants.

cmd/editorconfig-checker/main.go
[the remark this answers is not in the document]
klaernie: I actually now took a different route - it turns out `.NewConfig` does not need to produce errors at all, so I removed the error from it's return values.

klaernie:
I again incorporated all the feedback, I think this is the time for the final review now, unless something unexpected turns up

cmd/editorconfig-checker/main.go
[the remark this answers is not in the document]
ccoVeille: Thanks

cmd/editorconfig-checker/main_test.go
```diff
@@ -0,0 +1,161 @@
...
+	"bytes"
+	"os"
+	"runtime"
+	"strings"
+	"testing"
+)
+
+var mainHasRun chan int
+
+func captureReturncode(code int) {
```
ccoVeille, line 13:
```suggestion
func captureReturnCode(code int) {
```

cmd/editorconfig-checker/main_test.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+}
+
+func TestMainLoadingAncientConfig(t *testing.T) {
+	output, lastSeenCode := runWithArguments(t, "--debug", "--verbose", "--config", "testdata/ancient-config.json")
+	if lastSeenCode != exitCodeErrorOccurred {
+		t.Errorf("main exited with return code %d, but we expected %d", lastSeenCode, exitCodeErrorOccurred)
+	}
+	if !strings.Contains(output, "SpacesAftertabs") {
+		t.Errorf("main did not produce a warning that SpacesAftertabs is deprecated\nOutput:\n%s", output)
```
ccoVeille, line 87: I'm surprised it's SpaceAftertabs and not SpaceAfterTabs

But I assume it's out of the PR scope

ccoVeille (approved):

cmd/editorconfig-checker/main_test.go
klaernie: we're testing that  the deprecation warning is output, not the presence of the new option.

theoludwig (approved):

cmd/editorconfig-checker/main_test.go
klaernie: hm.. I never knew I treated "returncode" as a single word.. although I guess proper english would write that as "return code"... Anyway, fixed it.

klaernie:
I think if nothing else comes up, we should start the 72h now ;)

PR #381, opened 2024-09-16
chore: add a pre-commit config

klaernie:
If any developer uses [pre-commit](https://pre-commit.com/) it reliably ensures that writing non-conventional commit messages do not happen, and that the testsuite passes.

~I also threw in two goodies of preventing TODO and FIXME occurances in newly added lines (and yes, I had to `git commit --no-verify` to actually make this commit).~

One word of caveat: these are of cause only local tests, and are not enforcable in any way, since they run on the committers machine. Enforcement still needs to happen in the CI.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -10,17 +10,17 @@ repos:
         args: []
   - repo: local
     hooks:
-      - id: no-FIXME
-        name: check for FIXME
-        description: 'Detect lines containing FIXME'
-        entry: 'FIXME'
+      - id: no-fixmes
+        name: check for fixmes
+        description: 'Detect lines containing uppercase fixme'
+        entry: '[F][I][X][M][E]'
         language: pygrep
         stages:
           - pre-commit
-      - id: no-TODO
-        name: check for TODO
-        description: 'Detect lines containing TODO'
-        entry: 'TODO'
+      - id: no-todo
+        name: check for todos
+        description: 'Detect lines containing uppercase todo'
+        entry: '[T][O][D][O]'
         language: pygrep
         stages:
           - pre-commit
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,6 +8,10 @@ repos:
         stages:
           - commit-msg
         args: []
+  - repo: https://github.com/pre-commit/pre-commit.git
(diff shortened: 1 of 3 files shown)


mstruebing (commented):
Maybe we could also add a `make build` in there as well to make sure it builds - not sure if it is worth it.

klaernie:
We could - but I would argue that `make test` already builds the code - since that is required to run the tests against it. And it even covers a little bit more, since in a `go build ./...` all the test case code is not loaded at all.

ccoVeille:
While this is a good thing for people contributing to the project. The number people contributing to the project will stay limited.

We should consider adding the configuration file needed to use editorconfig-checker via pre-commit.

Of course, this could be addressed in another PR and another issue

.pre-commit-config.yaml
```diff
@@ -0,0 +1,34 @@
...
+      - id: conventional-pre-commit
+        stages:
+          - commit-msg
+        args: []
+  - repo: local
+    hooks:
+      - id: no-fixmes
+        name: check for fixmes
+        description: 'Detect lines containing uppercase fixme'
+        entry: '[F][I][X][M][E]'
```
ccoVeille, line 16: Maybe q comment about this strange format

Also, this would be enough, no?

```suggestion
        entry: '[F]IXME' 
```

ccoVeille:
Thanks for implementing the precommit hook config.

We will have to document it as an alternative way to use/install it

.pre-commit-config.yaml
klaernie: maybe I should just keep these two hooks out of that config - I added them to force myself to fix an issue by marking it with `// FIXME`, but preventing me from accidentally committing that marker. Maybe the better solution would be to allow committing such a marker, but add a blocker in the CI that refuses to pass when new FIXME or TODO items are added. That way one can commit a stub, mark it as to be fixed before merging, but discuss the interfaces around that code in the PR.

klaernie:
I now removed the fixme and todo grep hooks, and added a piece of documentation.

theoludwig (approved):
Why not, if it simplifies contributions to the project, and it is optional. :+1:

mstruebing (approved):

ccoVeille (approved):

PR #386, opened 2024-10-05
fix: typo in config, `SpacesAftertabs` => `SpacesAfterTabs`

hituzi-no-sippo:
#247 

> I think we should rather change this in the code to be SpacesAfterTabs as this is the more accurate name.
Would you be willing to provide a PR for this too?


klaernie (changes requested):
Since this is a breaking change we should provide a transition period in which we accept the old name in the config file, but warn that this property has been renamed.

pkg/config/config.go
```diff
@@ -204,7 +205,13 @@ func (c *Config) Merge(config Config) {
 	}
 
 	if config.SpacesAftertabs {
-		c.SpacesAftertabs = config.SpacesAftertabs
+		c.Logger.Warning("The configuration key `SpacesAftertabs` is deprecated. Use `SpacesAfterTabs` instead.")
+
+		c.SpacesAfterTabs = config.SpacesAftertabs
+	}
```
hituzi-no-sippo, line 211: * Should the value of `config.SpacesAftertabs` be stored in `c.SpacesAftertabs` ?
  Should I change the code to the following?
  ``` go
  if config.SpacesAftertabs {
  	c.Logger.Warning("The configuration key `SpacesAftertabs` is deprecated. Use `SpacesAfterTabs` instead.")

  	c.SpacesAftertabs = config.SpacesAftertabs
  	c.SpacesAfterTabs = config.SpacesAftertabs
  }
  ```

hituzi-no-sippo: * How can I make `editorconfig-checker` output a warning even if `SpacesAftertabs` is `false` ?
  The current code does't output the warning if `"SpacesAftertabs": false` is set in the configuration file.

  ``` console
  # "SpacesAftertabs": true
  $ cat .ecrc
  {
    "SpacesAftertabs": true
  }
  $ bin/ec
  The configuration key `SpacesAftertabs` is deprecated. Use `SpacesAfterTabs` instead.

  # "SpacesAftertabs": false
  $ sed -i 's/true/false/' .ecrc
  $ cat .ecrc
  {
    "SpacesAftertabs": false
  }
  $ bin/ec # No warning is output.
  ```

hituzi-no-sippo:
Thanks for your review.

I have the above two questions.
Would you mind answering them if you don't mind?
I am a beginner in Golang. So I don't know what to do about the above two points.

pkg/config/config.go
klaernie: I'm also fairly new to golang, so I just googled a bit and found this, which looks promising: https://stackoverflow.com/questions/31989595/detect-if-json-file-has-field

It should then mean, that after the config file has been decided you can check the `SpacesAftertabs` for not being a null pointer.

hituzi-no-sippo: Thanks for the help.
It worked.

theoludwig (approved):
Thank you!

It looks good to me. :smile:

klaernie (approved):

klaernie:
As much as it pains me (because I will have to rebase something ;) ) it's a good change to have! Thanks for implementing, @hituzi-no-sippo!

nietras:
This change breaks my build using Super Linter and I don't know how to fix given I don't have SpacesAftertabs in the file 😅

https://github.com/nietras/Sep/actions/runs/15208581410/job/42777265663

PR #389, opened 2024-10-11
fix: add `.ecrc` deprecation warning

### LINKED ISSUE #388: Deprecate by a warning the `.ecrc` configuration file
In #375, a new configuration file has been added: `.editorconfig-checker.json`.

In v4.0.0, we plan to drop the naming `ec` to use `editorconfig-checker` (#371), so we thought that we should deprecate usage of `.ecrc`, so that we can remove it in v4.0.0.

However, currently it is only deprecated in the `README.md`, we should **deprecate it at runtime**, by emitting a warning so that it will be easier for users to upgrade.

Similarly, it has been suggested to emit warnings in #386.

abuzaforfagun:
Fixes #388 

> In https://github.com/editorconfig-checker/editorconfig-checker/pull/375, a new configuration file has been added: .editorconfig-checker.json.
> 
> In v4.0.0, we plan to drop the naming ec to use editorconfig-checker (https://github.com/editorconfig-checker/editorconfig-checker/pull/371), so we thought that we should deprecate usage of .ecrc, so that we can remove it in v4.0.0.
> 
> However, currently it is only deprecated in the README.md, we should deprecate it at runtime, by emitting a warning so that it will be easier for users to upgrade.
> 
> Similarly, it has been suggested to emit warnings in https://github.com/editorconfig-checker/editorconfig-checker/pull/386.
>

### DIFF
--- a/cmd/editorconfig-checker/main.go
+++ b/cmd/editorconfig-checker/main.go
@@ -61,7 +61,6 @@ func init() {
 		configPaths = append(configPaths, defaultConfigFileNames[:]...)
 	} else {
 		configPaths = append(configPaths, configFilePath)
-
 	}
 
 	currentConfig, _ = config.NewConfig(configPaths)
--- a/cmd/editorconfig-checker/main.go
+++ b/cmd/editorconfig-checker/main.go
@@ -5,6 +5,7 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"strings"
 
 	"github.com/editorconfig-checker/editorconfig-checker/v3/pkg/config"
 	"github.com/editorconfig-checker/editorconfig-checker/v3/pkg/error"
@@ -60,12 +61,14 @@ func init() {
 		configPaths = append(configPaths, defaultConfigFileNames[:]...)
 	} else {
 		configPaths = append(configPaths, configFilePath)
-		if configFilePath == ".ecrc" {
-			logger.Warning("The default configuration file name `.ecrc` is deprecated. Use `.editorconfig-checker.json` instead.")
-		}
+
 	}
 
 	currentConfig, _ = config.NewConfig(configPaths)
+
+	if strings.Contains(currentConfig.Path, ".ecrc") {
+		logger.Warning("The default configuration file name `.ecrc` is deprecated. Use `.editorconfig-checker.json` instead.  You can simply rename it")
+	}
 
 	if init {
 		err := currentConfig.Save(version)
--- a/cmd/editorconfig-checker/main.go
+++ b/cmd/editorconfig-checker/main.go
@@ -67,7 +67,7 @@ func init() {
 	currentConfig, _ = config.NewConfig(configPaths)
 
 	if strings.Contains(currentConfig.Path, ".ecrc") {
-		logger.Warning("The default configuration file name `.ecrc` is deprecated. Use `.editorconfig-checker.json` instead.  You can simply rename it")
+		logger.Warning("The default configuration file name `.ecrc` is deprecated. Use `.editorconfig-checker.json` instead. You can simply rename it")
 	}
 
 	if init {
--- a/cmd/editorconfig-checker/main.go
+++ b/cmd/editorconfig-checker/main.go
@@ -60,6 +60,9 @@ func init() {
 		configPaths = append(configPaths, defaultConfigFileNames[:]...)
 	} else {
 		configPaths = append(configPaths, configFilePath)
+		if configFilePath == ".ecrc" {
+			logger.Warning("The default configuration file name `.ecrc` is deprecated. Use `.editorconfig-checker.json` instead.")
+		}
 	}
 
 	currentConfig, _ = config.NewConfig(configPaths)
--- a/cmd/editorconfig-checker/main.go
+++ b/cmd/editorconfig-checker/main.go
@@ -65,7 +65,7 @@ func init() {
 
 	currentConfig, _ = config.NewConfig(configPaths)
 
-	if strings.Contains(currentConfig.Path, ".ecrc") {
+	if strings.HasSuffix(currentConfig.Path, ".ecrc") {
 		logger.Warning("The default configuration file name `.ecrc` is deprecated. Use `.editorconfig-checker.json` instead. You can simply rename it")
 	}
 


cmd/editorconfig-checker/main.go
```diff
@@ -60,6 +60,9 @@ func init() {
 		configPaths = append(configPaths, defaultConfigFileNames[:]...)
 	} else {
 		configPaths = append(configPaths, configFilePath)
+		if configFilePath == ".ecrc" {
+			logger.Warning("The default configuration file name `.ecrc` is deprecated. Use `.editorconfig-checker.json` instead.")
```
ccoVeille, line 64:
```suggestion
			logger.Warning("The configuration file name `.ecrc` is deprecated. Use `.editorconfig-checker.json` instead. You can simply rename it")
```

cmd/editorconfig-checker/main.go
```diff
@@ -60,6 +60,9 @@ func init() {
 		configPaths = append(configPaths, defaultConfigFileNames[:]...)
 	} else {
 		configPaths = append(configPaths, configFilePath)
+		if configFilePath == ".ecrc" {
```
ccoVeille, line 63: I don't think the if statement is at the right place:

- here you are testing the content parameter `--config`
- it could be a relative or full path, so testing with an equality looks insufficient
- if someone passed a config path, even if it's the old name, we should accept silently.

ccoVeille (changes requested):

cmd/editorconfig-checker/main.go
```diff
@@ -60,6 +60,9 @@ func init() {
...
+			logger.Warning("The default configuration file name `.ecrc` is deprecated. Use `.editorconfig-checker.json` instead.")
+		}
 	}
 
 	currentConfig, _ = config.NewConfig(configPaths)
```
ccoVeille, line 68: A better test could be to check here for the currentConfig file name to contain `.ecrc` only if comfigFilePath is empty, then display the warning message

cmd/editorconfig-checker/main.go
ccoVeille: Here is a suggestion

https://github.com/editorconfig-checker/editorconfig-checker/pull/389#discussion_r1797617246

abuzaforfagun: @ccoVeille I have updated the code accordingly.

theoludwig (approved):
Thank you. :smile:

LGTM, with a green CI.

cmd/editorconfig-checker/main.go
```diff
@@ -60,10 +61,15 @@ func init() {
 		configPaths = append(configPaths, defaultConfigFileNames[:]...)
 	} else {
 		configPaths = append(configPaths, configFilePath)
+
 	}
 
 	currentConfig, _ = config.NewConfig(configPaths)
 
+	if strings.Contains(currentConfig.Path, ".ecrc") {
```
ccoVeille, line 69: Please look at

https://github.com/editorconfig-checker/editorconfig-checker/pull/389#discussion_r1797617246

```suggestion
	if configFile Path == "" && strings.Contains(currentConfig.Path, ".ecrc") {
```

Also, please consider using hassuffix

```suggestion
	if configFile Path == "" && strings.HasSuffix(currentConfig.Path, ".ecrc") {
```

theoludwig:
I think, that you will need to `rebase` and `git push --force` to have a green CI.

klaernie:
@abuzaforfagun : Why are all commits marked as breaking changes?

theoludwig:
> @abuzaforfagun : Why are all commits marked as breaking changes?

It will be fine, we will be able to "Squash and Merge" with the PR title as the commit message.
But yes, this change is not a breaking change, and the whole purpose of this change is to make the future planned breaking change less "painful".

theoludwig (approved):

cmd/editorconfig-checker/main.go
```diff
@@ -64,6 +65,10 @@ func init() {
 
 	currentConfig, _ = config.NewConfig(configPaths)
 
+	if strings.HasSuffix(currentConfig.Path, ".ecrc") {
```
ccoVeille, line 68: I ask everyone to consider doing this

```suggestion
	if configFilePath == "" && strings.HasSuffix(currentConfig.Path, ".ecrc") {
```

This way we won't report a problem to user when they provide a path with .ecrc with --config

The warning added via this feature is about asking the use to rename the config file. For me we should limit ourselves to when no --config is provided. Warning them in such case would be a problem.

cmd/editorconfig-checker/main.go
```diff
@@ -64,6 +65,10 @@ func init() {
 
 	currentConfig, _ = config.NewConfig(configPaths)
```
ccoVeille, line 66: Side note, out of this PR consideration

@theoludwig @klaernie why are we ignoring the error here?

ccoVeille: Same question for `config.Parse()` call a few liner below

cmd/editorconfig-checker/main.go
ccoVeille: I reopened the comment

Because I didn't see a fix for the point I raised

I commented here

https://github.com/editorconfig-checker/editorconfig-checker/pull/389#discussion_r1798291169

ccoVeille (changes requested):

cmd/editorconfig-checker/main.go
klaernie: Hm.. we could declare this a conscious decision - after all that is to encourage change to the new name. And some tooling like super-linter will automatically add a `--config` when they find a matching find in their directory (in this example `.github/linters/`

cmd/editorconfig-checker/main.go
klaernie: I've asked myself the same question for #380 and rewrote a bit of it, didn't I? But no matter, that should be a separate issue.

ccoVeille: Yes, exactly

cmd/editorconfig-checker/main.go
theoludwig: > encourage change to the new name

So if we take the decision that it's on purpose, that we also warn if the user provide `--config .ecrc`, to encourage using another name, are your (@ccoVeille) reviews blocking? Is this PR ready to be merged as is?

ccoVeille: Non blocking.

ccoVeille (approved):

cmd/editorconfig-checker/main.go
klaernie: Just read over it again, to be sure: in #380 I started verifying the error returned by `config.Parse()`, since that is the relevant point where reading a custom config file must succeed for us to continue, yet we can safely ignore the error if we tried to parse the default configuration file.

klaernie (approved):
Looks good now!

PR #392, opened 2024-10-13
chore: rename our own .ecrc to the new name

klaernie:
Seems like we totally forgot to actually rename our own config to the new name, so here that change is.


klaernie:
oh, that linter is not happy.. I'll have to fix that

klaernie:
Now everything is happy again!

ccoVeille (approved):

theoludwig (approved):
Ah, yes, we should not use our own deprecated thing. :laugh:

PR #414, opened 2025-01-08
chore: use go1.23(.4) to build and release

klaernie:
This is entirely a forward-looking change to not be in a haste to upgrade to a newer version.

And looking at the CI there seems nothing against it functionally.
Organizationally one could argue that we are forcing downstream consumers to upgrade to go1.23, but this would only be meaningful if they would consume parts of our codebase as a library, which I would not assume.


ccoVeille (approved):

theoludwig (approved):
Yes, let's upgrade Go, I don't think it will have any impact for our users, and that way we'll be able to merge #401.

mstruebing (approved):

PR #416, opened 2025-01-10
fix: provide both .tar.gz as well as .zip archives

klaernie:


klaernie:
I'm going to merge this once CI passes and run the release for it, since technically right now only the javascript integration is fixed to support zip archives.

Issue #413, opened 2025-01-08
VERSION file not updated at the root of the repository


theoludwig:
I don't know if we use this file for something or why it is there, but we might remove it, if it's not necessary: https://github.com/editorconfig-checker/editorconfig-checker/blob/main/VERSION

or if it's needed, find a way to update it with Release Please.

cs278:
https://github.com/editorconfig-checker/editorconfig-checker/blob/ec7219a7f1d6cda5ebb56fc5806a9486b084b49d/Dockerfile#L9

klaernie:
Technically we do no longer need it. It is still referenced in the old Dockerfile and Makefile, but it's content is not used for anything productive.

PR #418, opened 2025-01-12
chore: remove old build-binary-in-docker Dockerfile

klaernie:
The old Dockerfile is not really in use any more, and is definitely partially out of date. So rather than keeping this file around, it should be better to just drop it, and replace it with the one used by goreleaser.

This was brought to my attention in https://github.com/editorconfig-checker/editorconfig-checker/pull/371#discussion_r1912238701, so instead of just removing the outdated variable, it is better to remove the entire file.


Dockerfile
```diff
@@ -1,20 +1,17 @@
...
-RUN git config --global --add safe.directory /check
+RUN apk add --no-cache git \
+    && git config --global --add safe.directory /check
 WORKDIR /check/
-COPY --from=build /ec/bin/ec /usr/bin
 
-CMD ["ec"]
+COPY editorconfig-checker /usr/bin/
+
+# reasonale: I've see repackagers use COPY --from editorconfig-checker /usr/bin/ec /usr/bin/editorconfig-checker
```
ccoVeille, line 13: reasonale doesn't sound like a real word

ccoVeille (approved):

Dockerfile
theoludwig: Yeah + personally, I would avoid using "me", "I" in a comment.

theoludwig:
Not blocking, but is it possible to also remove the `VERSION` file also? So we can fix this issue #413

ccoVeille (approved):

theoludwig (approved):
Thanks!

PR #421, opened 2025-01-13
chore: replace DOCKER_USERNAME secret by a variable

klaernie:
I personally find it annoying that this public piece of information is redacted in the action output - GitHub tries to be on the safe side, so it replaces every occurence of the value (`mstreubing`) everywhere in the output. But since the value is not actually a secret value, it is better to replace it by a variable.

PS: the variable is already created, so nothing to do from this end.


mstruebing (approved):
As the variable and secret are exactly the same in this case, would it still censor the ~~secret~~ variable? We might need to delete it. 🤔 

Anyway, here is my approval

klaernie:
Correct, once this PR is merged we can remove the secret, since nothing uses it anymore. I think GitHub will not censor the value, if the secret is nowhere used in the workflow, but we should see as soon as it is merged.

theoludwig (approved):

mstruebing:
@klaernie do you have access rights to do that or do you need someone to assist you?

klaernie:
> @klaernie do you have access rights to do that or do you need someone to assist you?

I think managing org secrets is only allowed to org owners, which I'm not. Plus adding the renamed secret is something only @theoludwig can do anyway (since only he has access to the account currently)

PR #422, opened 2025-01-13
chore: replace token bound to a human by one bound to a bot

klaernie:
The automerge of dependabot is configured by a token bound to @mstruebing - which I find a bit confusing. I'd much rather like to use our bot @editorconfig-checker-bot, since it would rather reflect that the action was not done by a human.

On a separate note if we go for this, we maybe should rather save the token for the bot in an organization-level secret called `BOT_TOKEN`, and configure it to be used everywhere we currently are required to use a personal access token, since that means the management of the token is in one central point.

Afterwards we could remove these secrets on a repository level:
- `GH_TOKEN`
- `GH_TOKEN_PAT`

and on a org level:
- `GH_TOKEN_PAT`


.github/workflows/dependabot-automerge.yml
```diff
@@ -14,9 +14,9 @@ jobs:
...
         uses: dependabot/fetch-metadata@v2
         with:
-          github-token: "${{ secrets.GH_TOKEN_PAT }}"
+          github-token: "${{ secrets.RELEASE_PLEASE_TOKEN }}"
       - name: Enable auto-merge for Dependabot PRs
         run: gh pr merge --auto --squash "$PR_URL"
         env:
           PR_URL: ${{ github.event.pull_request.html_url }}
-          GITHUB_TOKEN: ${{ secrets.GH_TOKEN_PAT }}
+          GITHUB_TOKEN: ${{ secrets.RELEASE_PLEASE_TOKEN }}
```
klaernie, line 22:
```suggestion
          GITHUB_TOKEN: ${{ secrets.BOT_TOKEN }}
```

.github/workflows/dependabot-automerge.yml
```diff
@@ -14,9 +14,9 @@ jobs:
         id: metadata
         uses: dependabot/fetch-metadata@v2
         with:
-          github-token: "${{ secrets.GH_TOKEN_PAT }}"
+          github-token: "${{ secrets.RELEASE_PLEASE_TOKEN }}"
```
klaernie, line 17:
```suggestion
          github-token: "${{ secrets.BOT_TOKEN }}"
```

mstruebing (commented):
I've also automerge enabled in some other repos.
At least the action, I"m not entirely sure about the others. (https://github.com/editorconfig-checker/action-editorconfig-checker/blob/main/.github/workflows/dependabot-automerge.yaml#L22)

theoludwig (commented):
> On a separate note if we go for this, we maybe should rather save the token for the bot in an organization-level secret called `BOT_TOKEN`, and configure it to be used everywhere

Let's do that then, :smile:

klaernie:
@theoludwig I've rebased it so we could merge it as soon as the org-level secret exists.

theoludwig (approved):
I added `BOT_TOKEN` to the org secret.
Corresponding to the token of @editorconfig-checker-bot.

klaernie:
@theoludwig I think we need a second copy of that token in https://github.com/organizations/editorconfig-checker/settings/secrets/dependabot - else the pull requests made my dependabot are not allowed to use that token.

theoludwig:
> @theoludwig I think we need a second copy of that token in [github.com/organizations/editorconfig-checker/settings/secrets/dependabot](https://github.com/organizations/editorconfig-checker/settings/secrets/dependabot) - else the pull requests made my dependabot are not allowed to use that token.

:eyes: Okay, I added it. Thanks for the heads up.

klaernie:
> > @theoludwig I think we need a second copy of that token in [github.com/organizations/editorconfig-checker/settings/secrets/dependabot](https://github.com/organizations/editorconfig-checker/settings/secrets/dependabot) - else the pull requests made my dependabot are not allowed to use that token.
> 
> 👀 Okay, I added it. Thanks for the heads up.

Thanks, just tested it in the actions-ecc repo and it worked a treat!

PR #423, opened 2025-01-14
feat: only output "0 errors found" when verbose output is enabled

### LINKED ISSUE #420: Should we report "0 errors found" in a successful run?
> What I wanted to mention is that it would be neat to be silent if no errors are found. Right now you get a red "0 errors > found" after this PR - it's information one cannot act on, and thus could be skipped.

_Originally posted by @andreineculau in https://github.com/editorconfig-checker/editorconfig-checker/issues/360#issuecomment-2588036905_

klaernie:
- [x] write tests for PrintErrorCount()


klaernie:
now we also have tests to ensure this works correctly everytime. Please read it carefully, I'm a tad too sleepy to write code ;)

pkg/error/__snapshots__/error_test.snap
```diff
@@ -0,0 +1,22 @@
...
+
+---
+
+[TestPrintErrorCount/nonverbose-zero - 1]
+
+---
+
+[TestPrintErrorCount/verbose-ten - 1]
+[31;1m
+10 errors found[33;0m
```
ccoVeille, line 14: Could the --no-color mode be tested too?

Nothing test it for now and here could be a good place

klaernie: I think this should be it's own test in the logger package

klaernie: We actually have testing of the uncolorized mode:
https://github.com/editorconfig-checker/editorconfig-checker/blob/main/pkg/logger/logger_test.go#L56-L59
https://github.com/editorconfig-checker/editorconfig-checker/blob/main/pkg/logger/logger_test.go#L76-L79

mstruebing (approved):
LGTM

theoludwig (approved):
This commit is probably a `fix`, not a `build` because it is something releaseable.

In the Release Please they are describing the commits "releaseable": https://github.com/googleapis/release-please#how-should-i-write-my-commits, it's basically `feat`, `fix`, `perf`, or breaking change indicator like `!`, and it is often the same convention used by all the tools that use conventional commits.

klaernie:
Hm.. "fix" doesn't seem right either - I would lean more towards "feat"

theoludwig:
> doesn't seem right either - I would lean more towards "feat"

image

Let's use `feat` then. :smile:

theoludwig:
I just checked, our current semver policy states the following (ref: https://github.com/editorconfig-checker/editorconfig-checker#semantic-versioning-policy):
> Patch release (1.0.x -> 1.0.y)
> - Updates to output formats (error messages, logs, ...).

So might be a `fix` then, as `feat` means it's a minor release, and a bug `fix` is a patch release.
But honestly, that's not too important, as long as we correctly ensure semver major and breaking changes are correctly handled, and that we correctly differentiate commits that are releasable and the others that are not like this one: https://github.com/editorconfig-checker/editorconfig-checker/commit/e8bcbbc1a5b7dd5c0b25eadf747bdab4017ff6fa, it's fine.

By "releasable", I mean have an impact on ours users.

klaernie:
Hm.. I would think that this falls under "any new feature" for a minor release.. but that feature is almost not a feature, since it is so little..

PR #425, opened 2025-01-16
ci: use new goreleaser formats to get the archives we need

klaernie:
This can only be merged once goreleaser/goreleaser#5455 is released officially


klaernie:
goreleaser 2.6 shipped quite a while ago, so this is now ready to be used

mstruebing (approved):

Issue #197, opened 2022-05-02
Error when running in gitlab ci pipeline


dpuscher:
When I try to run the editorconfig-checker in the ci pipeline on GitLab I get the following error. Locally it works without problems.

### Output:
```
$ yarn lint:editorconfig
yarn run v1.22.18
$ editorconfig-checker
panic: cannot parse the ini file ".editorconfig": normalization error: insert_final_newline=off is not an acceptable value. strconv.ParseBool: parsing "off": invalid syntax
goroutine 1 [running]:
github.com/editorconfig-checker/editorconfig-checker/pkg/validation.ValidateFile(0xc000b255ec, 0x76, 0x0, 0x7[40](https://[...removed...]/-/jobs/1187671#L40)5e9, 0x5, 0x0, 0x0, 0x0, 0x9bb348, 0x0, ...)
	/home/maex/projects/own/editorconfig-checker/editorconfig-checker/pkg/validation/validation.go:44 +0xcef
github.com/editorconfig-checker/editorconfig-checker/pkg/validation.ProcessValidation(0xc0003[42](https://[...removed...]/-/jobs/1187671#L42)000, 0x[43](https://[...removed...]/-/jobs/1187671#L43)4f, 0x4c00, 0x0, 0x7405e9, 0x5, 0x0, 0x0, 0x0, 0x9bb348, ...)
	/home/maex/projects/own/editorconfig-checker/editorconfig-checker/pkg/validation/validation.go:176 +0x1a9
main.main()
	/home/maex/projects/own/editorconfig-checker/editorconfig-checker/cmd/editorconfig-checker/main.go:122 +0x426
error Command failed with exit code 2.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

### Editorconfig file:
```
root = true

[*]
charset = utf-8
end_of_line = lf
indent_size = 2
indent_style = space
insert_final_newline = true
trim_trailing_whitespace = true
```

Any idea what could be a fix here? Or is this a bug with the editorconfig-checker?

mstruebing:
The error says: 

`panic: cannot parse the ini file ".editorconfig": normalization error: insert_final_newline=off is not an acceptable value. strconv.ParseBool: parsing "off": invalid syntax`

That means you have a wrong value for `insert_final_newline` i.e. something like `insert_final_newline = off` instead of `false`. 
Could you try running with verbose and debug output enabled?
[...]

PR #427, opened 2025-01-18
fix: improve default excludes

klaernie:
After reading through past issues, I came to the conclusion that we should exclude both `.git` (#268 ) and `node_modules` (#197) by default.


ccoVeille (approved):

ccoVeille:
Another common exclude is "vendor" I think

theoludwig (approved):

PR #429, opened 2025-01-20
feat: add support for env var NO_COLOR

### LINKED ISSUE #280: Respect the `NO_COLOR` environment variable to disable default color output
There is an informal standard described at https://no-color.org/ aimed at command-line programs that produce color output by default to respect the `NO_COLOR` environment variable to disable default color output. It would be nice if `editorconfig-checker` followed it too. It does use colors by default, and there is the `-no-color` command-line option to disable that, but I see the benefit of joining other numerous command-line tools listed at https://no-color.org/#software-directly-supporting-no_color-to-disable-color-output-by-default so that users can easily disable color in many applications at once.

One small problem in context of `editorconfig-checker` - since the color output is default, there is only the `-no-color` flag to turn it off, but there isn't an option to turn color on. For programs supporting `NO_COLOR`, it seems to be expected, see https://no-color.org/#frequently-asked-questions:

> 2\. **How should configuration files and command-line arguments be processed in the presence of `$NO_COLOR`?**
> 
> User-level configuration files and per-instance command-line arguments should override `$NO_COLOR`. _A user should be able to export `$NO_COLOR` in their shell configuration file as a default, but configure a specific program in its configuration file to specifically enable color._

So I think such option allowing to override `NO_COLOR` setting and turn colors back on should be added - if there is already `-no-color` to disable colors, maybe `-color` could enable them?

klaernie:


cmd/editorconfig-checker/main.go
```diff
@@ -47,6 +47,15 @@ var cmdlineExclude string
...
 var writeConfigFile bool
 
+func enableNoColor(s string) error {
+	cmdlineConfig.NoColor = true
+	return nil
+}
+func disableNoColor(s string) error {
+	cmdlineConfig.NoColor = false
+	return nil
+}
```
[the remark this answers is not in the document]
klaernie: oh, neat that one can do that! TIL!

README.md
```diff
@@ -127,7 +127,9 @@ USAGE:
   -init
         creates an initial configuration
   -no-color
-        dont print colors
+        disables printing color
+  -color
```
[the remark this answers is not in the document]
klaernie: It feel to me this is too hardcoded - especially in existing  `.editorconfig-checker.json` configs - where we would need to transition to a different type or turn over everything relating to the existing no-color implementation.

Do you think turning it over entirely and deprecating the NoColor setting and introducing a new tristate string-typed config option would be better?

cmd/editorconfig-checker/main.go
```diff
@@ -78,6 +88,11 @@ func parseArguments() {
 	cmdlineConfig = config.Config{}
 	writeConfigFile = false
 
+	// check the NO_COLOR environment variable before parsing the arguments, so the arguments can override
+	if os.Getenv("NO_COLOR") != "" {
```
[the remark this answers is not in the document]
klaernie: I was treating it maybe too correctly according to their spec (emphasis mine):

> Command-line software which adds ANSI color to its output by default should check for a NO_COLOR environment variable that, when present and not an empty string (**regardless of its value**), prevents the addition of ANSI color.

But I changed that to at least ignore a set `NO_COLOR=0` environment variable (`\0` is just the c-style implementation for an empty nul-byte terminated string)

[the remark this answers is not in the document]
klaernie: I also added a case-insensitive comparison to `false`, since I assume someone will treat it as a boolean. But we can wait for the PR requesting `off` as an acceptable value ;)

README.md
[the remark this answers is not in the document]
ccoVeille: I didn't thought about the JSON config indeed

Please note I would expect --color to be equivalent to --color=always unless NO_COLOR env is set

cmd/editorconfig-checker/main.go
```diff
@@ -78,6 +88,11 @@ func parseArguments() {
 	cmdlineConfig = config.Config{}
 	writeConfigFile = false
 
+	// check the NO_COLOR environment variable before parsing the arguments, so the arguments can override
+	if nocolor := os.Getenv("NO_COLOR"); nocolor != "" && nocolor != "0" && !strings.EqualFold(nocolor, "false") {
```
ccoVeille, line 92: You should consider using strconv.ParseBool it supports true,/false,0/1, t/f

klaernie: good point, that makes that `if` a little bigger, but takes out a lot of guesswork.

cmd/editorconfig-checker/main.go
```diff
@@ -78,6 +89,17 @@ func parseArguments() {
...
+	// check the NO_COLOR environment variable before parsing the arguments, so the arguments can override
+	if nocolor := os.Getenv("NO_COLOR"); nocolor != "" {
+		if nocolorParsedAsBool, err := strconv.ParseBool(nocolor); err == nil {
+			if nocolorParsedAsBool {
+				enableNoColor("")
+			}
+		} else {
+			enableNoColor("")
+		}
+	}
```
ccoVeille, line 101: This would be clearer I think

```suggestion
	// check the NO_COLOR environment variable before parsing the arguments, so the arguments can override
	if nocolor := os.Getenv("NO_COLOR"); nocolor != "" {
		nocolorParsedAsBool, err := strconv.ParseBool(nocolor)
		if err != nil {
			// let's assume assume the user wanted the NO_COLOR
			nocolorParsedAsBool = true
		}
		if nocolorParsedAsBool {
			enableNoColor("")
		}
	}
```

cmd/editorconfig-checker/main_test.go
```diff
@@ -133,6 +135,68 @@ func TestMainDryRun(t *testing.T) {
...
+	}{
+		{"no-envvar-no-arg", env{}, args{}},
+		{"envvar-no-arg", env{"NO_COLOR": "1"}, args{}},
+		{"no-envvar-color-off", env{}, args{"--no-color"}},
+		{"no-envvar-color-on", env{}, args{"--color"}},
+		{"envvar-color-off", env{"NO_COLOR": "1"}, args{"--no-color"}},
+		{"envvar-color-on", env{"NO_COLOR": "1"}, args{"--color"}},
+		{"no-envvar-color-offon", env{}, args{"--no-color", "--color"}},
+		{"no-envvar-color-onoffon", env{}, args{"--color", "--no-color", "--color"}},
+		{"envvar-true", env{"NO_COLOR": "true"}, args{}},
```
ccoVeille, line 179: add a test about NO_COLOR: "yes" and another to "no", you will validate they fall back to true

cmd/editorconfig-checker/main_test.go
```diff
@@ -133,6 +135,68 @@ func TestMainDryRun(t *testing.T) {
...
+		for name := range envs {
+			origValue, has := originalEnvs[name]
+			if has {
+				_ = os.Setenv(name, origValue)
+			} else {
+				_ = os.Unsetenv(name)
+			}
+		}
+	}
+}
```
ccoVeille, line 160: This should be preferred, the error you ignore will get caught, and it would avoid issue on testes ran in parallel 

```suggestion
func envSetter(t *testing.T, envs map[string]string) (closer func()) {
	originalEnvs := map[string]string{}

	for name, value := range envs {
		if originalValue, ok := os.LookupEnv(name); ok {
			originalEnvs[name] = originalValue
		}
		t.Setenv(name, value)
	}

	return func() {
		for name := range envs {
			origValue, has := originalEnvs[name]
			if has {
				t.Setenv(name, origValue)
			} else {
				t.Unsetenv(name)
			}
		}
	}
}
```

klaernie: hm.. looking at `t.Setenv` - this means we can thin the helper much more, since it automatically does the cleanup.

ccoVeille: That's its purpose yes

cmd/editorconfig-checker/main_test.go
klaernie: done

README.md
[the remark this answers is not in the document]
klaernie: I think implementing `--color={auto|never|always}` would in my mind also require detection if the current output destination actually supports color. For this PR I'd like to concentrate on the bare minimum - supporting the environment variable with correct precedence according to the spec.

cmd/editorconfig-checker/main_test.go
```diff
@@ -133,6 +135,52 @@ func TestMainDryRun(t *testing.T) {
...
+	// we use the error message of a missing config file to test the coloredness of the output
+	defaultArgs := []string{
+		`--exclude=""`, "--ignore-defaults",
+		"testdata/trailing-whitespace.txt",
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			for name, value := range test.env {
+				t.Setenv(name, value)
```
ccoVeille, line 175: Life changer, right?

cmd/editorconfig-checker/main_test.go
```diff
@@ -133,6 +135,52 @@ func TestMainDryRun(t *testing.T) {
...
+		// test that the last argument wins
+		{"no-envvar-color-offon", env{}, args{"--no-color", "--color"}},
+		{"no-envvar-color-onoffon", env{}, args{"--color", "--no-color", "--color"}},
+		// test that boolean values in the env var work correctly
+		{"envvar-true", env{"NO_COLOR": "true"}, args{}},
+		{"envvar-false", env{"NO_COLOR": "false"}, args{}},
+		{"envvar-zero", env{"NO_COLOR": "0"}, args{}},
+		// test that pure string values are correctly handled as setting NoColor
+		{"envvar-yes", env{"NO_COLOR": "yes"}, args{}},
+		{"envvar-no", env{"NO_COLOR": "no"}, args{}},
```
ccoVeille, line 163:
```suggestion
		// test that unsupported string values are handled as NoColor=1
		{"envvar-yes", env{"NO_COLOR": "yes"}, args{}},
		{"envvar-no", env{"NO_COLOR": "no"}, args{}},
		{"envvar-foo", env{"NO_COLOR": "foo"}, args{}},
```

I also added a test, that was missing I think

ccoVeille (approved):
One last minor change

But I can approve

cmd/editorconfig-checker/main_test.go
klaernie: indeed!

theoludwig (approved):
LGTM

cmd/editorconfig-checker/main_test.go
klaernie: I added a bit more ridiculous version of that. But essentially that last test case also just exercises the path that the values `yes` and ̀`no` would take.

README.md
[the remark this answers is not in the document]
ccoVeille: Do you agree if we open a new issue for this?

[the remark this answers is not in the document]
klaernie: Definitely!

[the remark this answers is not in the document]
ccoVeille: Here it is

- https://github.com/editorconfig-checker/editorconfig-checker/issues/433

PR #437, opened 2025-02-02
ci: run golangci-lint

### LINKED ISSUE #434: golangci-lint: enable usetesting
Following .you suggestion of using t.SetEnv

_Originally posted by @ccoVeille in https://github.com/editorconfig-checker/editorconfig-checker/pull/429#discussion_r1923314165_

We could activate `usetesting`

https://golangci-lint.run/usage/linters/#usetesting 

Available since	v1.63.0

https://github.com/ldez/usetesting

klaernie:
Since @ccoVeille mentioned it in #434, I thought we should give it a spin.


klaernie:
Hm, with the current configuration we are at 334 found issues - now I wonder if we should just accept the current issues and just disallow new ones, or refine the list of enabled linters. Quite a few ones are definitely helpful - we have for example many cases where we create new errors on the fly, which would be better solved by defining constants.

Do you have any cool default set to use, @ccoVeille ?

ccoVeille:
I do !

https://github.com/ccoVeille/golangci-lint-config-examples

Please also note, the settings you chose might seem pretty complete, while it isn't.
Many golangci-lint linters are using default rulesets that all pretty lax, and not strict.

There are a lot of things to enable in govet, gostrict, staticcheck, revive to get interesting settings. Please note, not all of them are interesting. Some are in forcing rules barely no one want. Such as "never reuses a variable for error, define new one".

Discovering, enabling, and configuring linters is a journey, as interesting and perilous than going to the Mordor. You can loose yourself.

That's why I created my project of examples of golangci-lint configuration examples.

BTW, I think I will have to update my README file with something close to what I just wrote.

ccoVeille:
Then to reply your questions about what could be the path to enable and fix the issues.

I would say the strategy depends on the time you want to spend on fixing the issues golangci-lint may report.

I would say it's important to also do not block contributions to project.

There are many way to achieve this.

Define a .golangci.yml pretty strict at the project root. This way IDE will report the issues. So core developers and advanced Go contributors who know golangci-lint will see what they need to fix.

Then you have multiple options:
- you define a relaxed golangci-lint configuration option for the CI, that can stay in .github folder.

So here, you have somehow to fix the issues reported by this configuration files before enforcing the CI to block them. Otherwise you block contributions.

- second option is to play with `golangci-lint run -new` that can be used locally, but also configured on the golangci-lint-action via the with field.

This feature is there to say "OK, old code sucks, we know that, but we want new code to be clean"

Everything is a matter of balance:
- the quality you want for your code.
- the amount of linters you want to use
- the amount of issues currently in the code base
- the amount of time you want to spend on fixing first batch of errors reported with a relaxed set of rules
- how much people will understand/agree the errors reported are qualitative and not noise, opinionated.

It seems to be complex, but the gain is real. Also, don't be afraid, it's a journey, time is needed, but there are multiple rewards in trying such a quest:
- working with golangci-lint rules was a huge opportunity for me. It learned a lot of valuable things in Go, but also as a developer/architect/team lead.
- I know a lot of these rules now. I write code that complies them by default now.
- the number of bugs I caught with the CI **before they were merged** is insane. People with less experience in Go, development skills, and unfortunately people who don't care about shipping "bad code that works", they will all face the same CI warning and blockers.
- the gain is important for code reviewers as the "obvious" errors are reported. So when you receive a PR, most little errors were caught and fixed before you review
- the gain is huge as a maintainer, it catches real bug that experienced reviewers can miss.

I talked a lot, I hope I didn't afraid you. It's a very interesting and fascinating topic.

I can help of course, even if I'm less active right now for personal reason.

klaernie:
I was hoping for such a good base configuration to get started - and 17 issues is much more manageable that over 300.

Fixing the first set of issues was quite easy, but now I'm a bit stuck on how to handle the remaining three:
```
kandre@mainframe(pts/18)[golangci-lint!] ~/git-repos/ecc/editorconfig-checker % golangci-lint run
pkg/validation/validators/validators.go:14:1: File is not properly formatted (gci)
        // x-release-please-end
^
pkg/validation/validators/validators_test.go:10:1: File is not properly formatted (gci)
        // x-release-please-end
^
pkg/config/config.go:16:1: File is not properly formatted (gci)
        // x-release-please-end
^
kandre@mainframe(pts/18)[golangci-lint] ~/git-repos/ecc/editorconfig-checker %
```

ccoVeille:
23 issues:
* errcheck: 2
* gofmt: 1
* goimports: 1
* misspell: 1
* revive: 6
* staticcheck: 9
* thelper: 1
* unused: 1
* usetesting: 1

It grew over the time, but it can be fixed I think. The following commands can help

golangci-lint fmt
go fix

klaernie:
This branch is now finally ready to be reviewed. Most of the issues were minor gotchas and improvements to readability - especially the De Moran's law fix took me an hour to understand what the intend of the condition was - and I was the one who wrote it.

I would highly recommend reviewing this PR commit by commit - looking at it all at once it quite a chunk to chew through.

theoludwig (approved):

klaernie:
I rebased the branch to fix the merge conflict in `.github/workflows/ci.yaml` (both the windows CI and this branch added new jobs to the bottom of the file)

klaernie:
and with the rebase came a few new violations, that were easily fixed

theoludwig (approved):

.gitignore
```diff
@@ -6,3 +6,5 @@ result-bin
 path
 node_modules
 dist/
+# vim swapfiles
+**/*.sw[pomn]
```
ccoVeille, line 10: This might be OK. But it doesn't sound related to this PR

pkg/fix/fix.go
```diff
@@ -17,12 +17,12 @@ import (
 	// x-release-please-end
 )
 
-// FixFile applies the safe, supported fixes for filePath. It returns true when
+// File applies the safe, supported fixes for filePath. It returns true when
```
ccoVeille, line 20: Here you are changing the name of an exported method. It's a bad things.

So here I feel you should add a //nolint, abd do not rename

pkg/validation/validators/validators.go
```diff
@@ -165,14 +166,14 @@ func MaxLineLength(line string, maxLineLength int, charSet string) error {
...
-		return fmt.Errorf("Line too long (%d instead of %d)", length, maxLineLength)
+		return fmt.Errorf("line too long (%d instead of %d)", length, maxLineLength)
 	}
 
 	return nil
 }
 
 // Charset validates a file's charset
-func Charset(charsetWanted string, charsetFound string, config config.Config) error {
+func Charset(charsetWanted string, charsetFound string) error {
```
ccoVeille, line 176: Here you are changing the function signature.

Maybe you will have to keep config.Config and do not use it

.gitignore
klaernie: yeah, that was entirely my frustration of the maybe 10th time failing the test suite, while I still had vim open in the background. But it wouldn't really warrant it's own PR either..

pkg/fix/fix.go
klaernie: I would not assume that anybody would be using this method by using our modules, which are technically public, but one could as easily argue that they could as well be declared internal.
Furthermore this particular module is in our main branch for not even three days.

IMHO the argument should be: what is the best name for the function, when calling it?

pkg/validation/validators/validators.go
klaernie: I don't see a reason to keep it part of the interface, there quite a few of the other validators don't take a `config.Config` argument (3 do, 5 don't). So I wouldn't declare it part of a pattern at the moment.

pkg/fix/fix.go
ccoVeille: Let's rename it then

pkg/validation/validators/validators.go
ccoVeille: It's a breaking changes as the signature of an exported method is altered, so I reported.

But I'm fine removing it

ccoVeille (approved):

pkg/fix/fix.go
theoludwig: Why was this method renamed? There was a linting error because of `FixFile` naming?

`FixFile` is a better name to describe what it does than just `File`. 😅

pkg/validation/validators/validators.go
theoludwig: I don't think anyone is using `editorconfig-checker` as a Go module, most people, if not everyone, is using it as a CLI.

If needed, might put a "rule"/info to say that internal code changes/function renames/signatures changed should not be considered as a breaking change: https://github.com/editorconfig-checker/editorconfig-checker#semantic-versioning-policy

pkg/fix/fix.go
klaernie: The linter complained about the doubling it was originally named `fix.FixFile()`, now it's `fix.File()`.

pkg/validation/validators/validators.go
klaernie: Maybe we should have a separate discussion about renaming dir `pkg` to `internal` - that would enforce this truly. https://go.dev/doc/go1.4#internalpackages

ccoVeille: Yes, moving to this should be considered

klaernie: I'll build this after merging this branch - luckily that's a rename and a `perl -pie`, and best we do this with the 4.0 release, since this is semantically a breaking change

pkg/fix/fix.go
ccoVeille: Follow up PR

- https://github.com/editorconfig-checker/editorconfig-checker/pull/619

PR #447, opened 2025-03-02
fix: check for exclusion before MIME type

### LINKED ISSUE #442: exclude panic, can't access excluded file
test on windows:
```
❯ ec -exclude "(build|vs2022|tmp)"                                                                                           
panic: open vs2022/.vs/vAmigaSDL/FileContentIndex/4e84aa57-8be4-4f30-ac08-340db9d4df3b.vsidx: The process cannot access the file because it is being used by another process.
```

klaernie:
Since the exclusion check is cheap and requires no external interaction, it is better to check for the exclusion before opening and reading the file to determine the MIME type.

Besides the marginal speedup this also fixes #442 (where exclusions are used to prevent editorconfig-checker from reading inaccessible files)


mstruebing (approved):
LGTM

theoludwig (approved):

PR #440, opened 2025-02-08
ci: extend the CI to MacOS to catch regressions

klaernie:

PR #451, opened 2025-03-07
ci: extend the CI to windows

klaernie:
Broken out of #440, since windows needs more work than MacOS, and we do not want failing CI in the main branch.


klaernie:
Oh wow, that was surprisingly easy to get running. I must have spent too much time in GitHub actions the last weeks.

theoludwig (approved):

PR #457, opened 2025-03-26
feat: Support chardet config file setting

### LINKED ISSUE #40: Support checking all files' text encodings, if declared to be UTF-8
Enforcing a specified text encoding in a monorepo rife with all kinds of textual resources is hard for crossplatform development teams. This feature would be of great value, and simple to implement. Especially if only a BOM-check is done. Alternatively, every text file that is covered by the EditorConfig settings can be decoded as UTF-8 on the fly.

rasa:
Fixes #40

### DIFF
--- a/pkg/encoding/encoding.go
+++ b/pkg/encoding/encoding.go
@@ -3,6 +3,7 @@ package encoding
 
 import (
 	"fmt"
+	"slices"
 	"sort"
 	"strings"
 	"unicode/utf8"
@@ -19,6 +20,23 @@ import (
 )
 
 const BinaryData = "binary"
+
+const (
+	// See https://spec.editorconfig.org/#supported-pairs
+	CharsetLatin1  = "latin1"
+	CharsetUTF8    = "utf-8"
+	CharsetUTF8BOM = "utf-8-bom"
+	CharsetUTF16BE = "utf-16be"
+	CharsetUTF16LE = "utf-16le"
+)
+
+var ValidCharsets = []string{
+	CharsetLatin1,
+	CharsetUTF8,
+	CharsetUTF8BOM,
+	CharsetUTF16BE,
+	CharsetUTF16LE,
+}
 
 var encodings = map[string]encoding.Encoding{
 	// In https://github.com/golang/text/blob/HEAD/encoding/htmlindex/map.go#L64 and
@@ -156,20 +174,32 @@ func decodeText(contentBytes []byte, charset string) (string, error) {
 	return string(contentBytes), nil
 }
 
-func getEncoding(charset string) (encoding.Encoding, bool) {
+func normalizeCharsetName(charset string) string {
 	r := strings.NewReplacer("-", "", "_", "")
-	key := strings.ToLower(r.Replace(charset))
+	return strings.ToLower(r.Replace(charset))
+}
+
+func getEncoding(charset string) (encoding.Encoding, bool) {
+	key := normalizeCharsetName(charset)
 	enc, ok := encodings[key]
 	return enc, ok
 }
 
 var binaryChars = [256]bool{}
 
+var normalizedCharsets []string
+
 func init() {
 	// Allow tab (9), lf (10), ff (12), and cr (13)
 	trues := []byte{0, 1, 2, 3, 4, 5, 6, 7, 8, 11, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25, 26, 27, 28, 29, 30, 31}
 	for _, i := range trues {
 		binaryChars[i] = true
+	}
+
+	normalizedCharsets = make([]string, len(ValidCharsets))
+
+	for i, charset := range ValidCharsets {
+		normalizedCharsets[i] = normalizeCharsetName(charset)
 	}
 }
 
@@ -181,4 +211,52 @@ func IsBinaryFile(rawFileContent []byte) bool {
 		}
 	}
 	return false
+}
+
+// IsValid checks charset is supported by the editorconfig spec.
+func IsValid(charset string) bool {
+	charset = normalizeCharsetName(charset)
+	return slices.Contains(normalizedCharsets, charset)
+}
+
+var configCharsetMap = map[string]string{
+	"iso88591": CharsetLatin1,
+	// These files are almost always iso88591, so we don't want to annoy users
+	// with false negatives.
+	"windows1252": CharsetLatin1,
+}
+
+// ToConfigCharset returns the encoding found in the config file that matches,
+// the file's actual encoing.
+func ToConfigCharset(charset string) string {
+	charset = normalizeCharsetName(charset)
+	if charset == "iso88591" {
+		return CharsetLatin1
+	}
+	return charset
+}
+
+// Equal checks the two charset names for equality
+func Equal(charsetFound, charsetWanted string, bom bool) bool {
+	charsetFound = normalizeCharsetName(charsetFound)
+	charsetWanted = normalizeCharsetName(charsetWanted)
+	latin1, ok := configCharsetMap[charsetFound]
+	if ok {
+		charsetFound = latin1
+	}
+	if bom {
+		charsetFound += "bom"
+	}
+
+	latin1, ok = configCharsetMap[charsetWanted]
+	if ok {
+		charsetWanted = latin1
+	}
+
+	// latin1 files are utf8 files, too.
+	if strings.HasPrefix(charsetWanted, "utf8") && charsetFound == CharsetLatin1 {
+		return true
+	}
+
+	return charsetFound == charsetWanted
 }
--- a/pkg/encoding/encoding_test.go
+++ b/pkg/encoding/encoding_test.go
@@ -183,3 +183,94 @@ func TestIsBinaryFile(t *testing.T) {
 		}
 	}
 }
+
+type isEqualTest struct {
+	filename string
+	charset  string // empty for invalid charset
+}
+
+var isEqualTests = []isEqualTest{
+	{"8859_1_da.html", "latin1"},
+	{"8859_1_de.html", "latin1"},
+	{"8859_1_en.html", "latin1"},
+	{"8859_1_es.html", "latin1"},
+	{"8859_1_fr.html", "latin1"},
+	{"8859_1_pt.html", "latin1"},
+	{"ascii.txt", "latin1"},
+	{"big5.html", ""},
+	{"candide-gb18030.txt", ""},
+	// This file fails the test, but it may not be a valid UTF-16LE file.
+	// For example, the Linux file command doesn't identify the file as
+	// "Unicode text, UTF-16, little-endian text"
+	// but simply
+	// "data"
+	// but since the file is from
+	// https://cs.opensource.google/go/x/text/+/master:encoding/testdata/
+	// I think it's correct to fail the test, and fix the chardet package.
+	// {"candide-utf-16le.txt", "utf16le"},
+	{"candide-utf-32be.txt", ""},
+	{"candide-utf-8.txt", "utf8"},
+	{"candide-windows-1252.txt", "latin1"},
+	{"cp865.txt", ""},
+	{"euc_jp.html", ""},
+	{"euc_kr.html", ""},
+	{"gb18030.html", ""},
+	{"html.html", "latin1"},
+	{"html.iso88591.html", "latin1"},
+	{"html.svg.html", "latin1"},
+	{"html.usascii.html", "latin1"},
+	{"html.utf8bomdetect.html", "utf8bom"},
+	{"html.utf8bom.html", "utf8bom"},
+	{"html.utf8bomws.html", "utf8bom"},
+	{"html.utf8.html", "latin1"}, // File is misnamed
+	{"html.withbr.html", "latin1"},
+	{"iso88591.txt", "latin1"},
+	{"koi8_r.txt", ""},
+	{"latin1.txt", "latin1"},
+	{"rashomon-euc-jp.txt", ""},
+	{"rashomon-iso-2022-jp.txt", ""}, // byte 89 is an Esc (ASCII 27)
+	{"rashomon-shift-jis.txt", ""},
+	{"rashomon-utf-8.txt", "utf8"},
+	{"shift_jis.html", ""},
+	{"sunzi-bingfa-gb-levels-1-and-2-hz-gb2312.txt", ""},
+	{"sunzi-bingfa-gb-levels-1-and-2-utf-8.txt", "utf8"},
+	{"sunzi-bingfa-simplified-gbk.txt", ""},
+	{"sunzi-bingfa-simplified-utf-8.txt", "utf8"},
+	{"sunzi-bingfa-traditional-big5.txt", ""},
+	{"sunzi-bingfa-traditional-utf-8.txt", "utf8"},
+	{"unsu-joh-eun-nal-euc-kr.txt", ""},
+	{"unsu-joh-eun-nal-utf-8.txt", "utf8"},
+	{"utf16bebom.txt", "utf16be"},
+	{"utf16lebom.txt", "utf16le"},
+	{"utf16.txt", "utf16le"},
+	{"utf32bebom.txt", ""},
+	{"utf32lebom.txt", ""},
+	{"utf8_bom.html", "utf8bom"},
+	{"utf8.html", "utf8"},
+	{"utf8-sdl.txt", "latin1"}, // should be utf8
+	{"utf8.txt", "utf8"},
+	{"utf8.txt-encoding-test-files.txt", "utf8"},
+}
+
+func TestEqual(t *testing.T) {
+	for _, tt := range isEqualTests {
+		filename := "testdata/" + tt.filename
+		rawFileContent, err := os.ReadFile(filename)
+		if err != nil {
+			panic(err)
+		}
+		_, charset, err := DecodeBytes(rawFileContent)
+		charset = normalizeCharsetName(charset)
+		if tt.charset == "" {
+			if IsValid(charset) {
+				t.Errorf("Equal(): %q: expected: %v, got: %v", tt.filename, tt.charset, charset)
+			}
+			continue
+		}
+		bom := tt.charset == "utf8bom"
+		// t.Logf("%q: %q %q %v", tt.filename, charset, tt.charset, bom)
+		if !Equal(charset, tt.charset, bom) {
+			t.Errorf("Equal(): %q: expected: %v, got: %v", tt.filename, tt.charset, charset)
+		}
+	}
+}
--- a/README.md
+++ b/README.md
@@ -48,10 +48,12 @@ Currently, implemented editorconfig features are:
 - `indent_style`
 - `indent_size`
 - `max_line_length`
+- `charset`
 
 Unsupported features are:
 
-- `charset`
+- `spelling_language`
+- `tab_width`
 
 ## Quickstart
 
@@ -103,6 +105,8 @@ USAGE:
         config
   -debug
         print debugging information
+  -disable-charset
+        disables the charset check
   -disable-end-of-line
         disables the trailing whitespace check
   -disable-indent-size
--- a/cmd/editorconfig-checker/main.go
+++ b/cmd/editorconfig-checker/main.go
@@ -78,6 +78,7 @@ func init() {
 	flag.BoolVar(&cmdlineConfig.Disable.Indentation, "disable-indentation", false, "disables the indentation check")
 	flag.BoolVar(&cmdlineConfig.Disable.IndentSize, "disable-indent-size", false, "disables only the indent-size check")
 	flag.BoolVar(&cmdlineConfig.Disable.MaxLineLength, "disable-max-line-length", false, "disables only the max-line-length check")
+	flag.BoolVar(&cmdlineConfig.Disable.Charset, "disable-charset", false, "disables only the charset check")
 }
 
 // parse the arguments from os.Args
--- a/cmd/editorconfig-checker/main_test.go
+++ b/cmd/editorconfig-checker/main_test.go
@@ -139,6 +139,9 @@ func TestMainColorSupport(t *testing.T) {
 	type env map[string]string
 	type args []string
 
+	// Otherwise the first test fails...
+	t.Setenv("NO_COLOR", "")
+
 	tests := []struct {
 		name string
 		env  env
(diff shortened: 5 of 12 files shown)


ccoVeille:
Please let me know when we can review, I have already remarks, but I want for your go, because I try to cure myself

rasa:
@ccoVeille Yeah, hold off for a bit. I'm gonna toss chardet and use our own code to determine if a file is latin1, utf-8, utf-8-bom, utf-16be or utf-16le.

Here's a start:
```
package main

import (
	"bytes"
	"fmt"
	"unicode/utf16"
	"unicode/utf8"
)

func detectEncoding(data []byte) string {
	if len(data) >= 2 {
		if bytes.HasPrefix(data, []byte{0xFE, 0xFF}) {
			return "utf-16be (with BOM)"
		}
		if bytes.HasPrefix(data, []byte{0xFF, 0xFE}) {
			return "utf-16le (with BOM)"
		}
	}

	if utf8.Valid(data) {
		return "utf-8"
	}

	if isValidUTF16LE(data) {
		return "utf-16le (no BOM)"
	}
	if isValidUTF16BE(data) {
		return "utf-16be (no BOM)"
	}

	if isLikelyLatin1(data) {
		return "latin1"
	}

	return "binary (unknown or invalid text)"
}

func isValidUTF16LE(data []byte) bool {
	if len(data)%2 != 0 {
		return false
	}
	u16 := make([]uint16, len(data)/2)
	for i := 0; i < len(u16); i++ {
		u16[i] = uint16(data[2*i]) | uint16(data[2*i+1])<<8
	}
	decoded := utf16.Decode(u16)
	for _, r := range decoded {
		if r == utf8.RuneError {
			return false
		}
	}
	return true
}

func isValidUTF16BE(data []byte) bool {
	if len(data)%2 != 0 {
		return false
	}
	u16 := make([]uint16, len(data)/2)
	for i := 0; i < len(u16); i++ {
		u16[i] = uint16(data[2*i+1]) | uint16(data[2*i])<<8
	}
	decoded := utf16.Decode(u16)
	for _, r := range decoded {
		if r == utf8.RuneError {
			return false
		}
	}
	return true
}

func isLikelyLatin1(data []byte) bool {
	for _, b := range data {
		// C0 controls (0x00–0x1F), except tab (0x09), LF (0x0A), FF (0x0C), CR (0x0D)
		if b < 0x20 {
			switch b {
			case 0x09, 0x0A, 0x0C, 0x0D:
				continue
			default:
				return false
			}
		}
		// C1 controls (0x80–0x9F) — generally disallowed in Latin-1 text
		if b >= 0x80 && b <= 0x9F {
			return false
		}
	}
	return true
}

func main() {
	data := []byte("Hello,\tworld!\n") // Replace with file or test content
	fmt.Println("Detected encoding:", detectEncoding(data))
}
```
See https://go.dev/play/p/9yJvV-P-O5k

Based on my research, I think we should interpret `latin1` to mean `iso8859-1`, so if a file has any bytes in the 00-31 (except tab, lf, cr, and ff) or 128-159 range (`windows-1252` uses these), we would reject it as not `latin1`. If the user wants to allow those bytes, they need to use `charset = unset`.

Alternatively, we treat `latin1` as `unset`, and allow any byte stream, including valid uft8/16/32.

Thoughts?

(love the pic btw!)

klaernie:
I would probably ignore the entire set of control characters entirely (0-32). If I recall correctly they always stayed the same in meaning throughout all the iso-8859 family and Unicode.

Also IIRC if we only find byte values <128 it could be both utf8 and latin1, so both should be accepted.

ccoVeille:
I looked for a lib about that.

More looking for examples and ideas, more than looking for a lib to import.

I found this

https://github.com/softlandia/cpd
License Apache 2.0

I didn't check about how they handle the char 0-32.

But, I can tell their test files are interesting.
 
I like it considers codepage I wouldn't have thought about.

Does anyone know another lib? We could look at.

rasa:
> I found this
> 
> https://github.com/softlandia/cpd License Apache 2.0

@ccoVeille Thank you for the suggestion. Unfortunately, of the ISO8859s, it only identifies ISO8859-5, not ISO8859-1, which is what I think is the best match for the `latin1` config setting.

rasa:
> I would probably ignore the entire set of control characters entirely (0-32). If I recall correctly they always stayed the same in meaning throughout all the iso-8859 family and Unicode.

@klaernie Sorry, I don't quite follow. The 0-32 characters (other than TAB, FF, LF, CR) are the best way to determine if a file is text, or binary. It's how dos2unix, and many other programs, determines this. And utf-16/32 uses 0-32, so I don't see how they have the "same meaning" when in a utf-8 file, or an iso-8859-1 file.
 
> Also IIRC if we only find byte values <128 it could be both utf8 and latin1, so both should be accepted.

That's true. The trick is to determine if a file is `latin1` (aka iso-8859-1), or some other non-UTF encoding. The solution is clearly non-trivial, and the best solution I've found so far is [uchardet](https://gitlab.freedesktop.org/uchardet/uchardet).

klaernie:
> > I would probably ignore the entire set of control characters entirely (0-32). If I recall correctly they always stayed the same in meaning throughout all the iso-8859 family and Unicode.
> 
> @klaernie Sorry, I don't quite follow. The 0-32 characters (other than TAB, FF, LF, CR) are the best way to determine if a file is text, or binary. It's how dos2unix, and many other programs, determines this. And utf-16/32 uses 0-32, so I don't see how they have the "same meaning" when in a utf-8 file, or an iso-8859-1 file.

I was thinking of the meaning of the first 32 unicode codepoints being identical in meaning to the first 32 in ASCII - but I didn't think about the fact that despite them being the first 32 codepoints they are not represented as bytes with the values 0-32 in utf16 and utf32. However in utf8 this assumption would hold.

But no matter, you are indeed correct that this is the only chance to differentiate text from binary files. I think I should no spend too much time on GitHub having just gotten out of bed before my first coffee ;)

Thanks a lot for the effort you are putting into this!

rasa:
The new chardet library works much better than the old one. Here are the only failures:

```
--- FAIL: TestEqual (1.39s)
    encoding_test.go:283: Equal(): "iso88591.txt": expected: latin1, got: windows1255
    encoding_test.go:105: result={Encoding:Windows-1255 Confidence:0.99 Language:Hebrew} (first 4 bytes: 54686973)
    encoding_test.go:283: Equal(): "utf8-sdl.txt": expected: latin1, got: windows1254
    encoding_test.go:105: result={Encoding:Windows-1254 Confidence:0.5250663680561672 Language:Turkish} (first 4 bytes: 5554462d)
    encoding_test.go:283: Equal(): "utf8.txt": expected: utf8, got: windows1254
    encoding_test.go:105: result={Encoding:Windows-1254 Confidence:0.5132700345166831 Language:Turkish} (first 4 bytes: 7072656d)
FAIL
```
I need to drill into these.

rasa:
So I added the 159 testdata files from 
https://gitlab.freedesktop.org/uchardet/uchardet/-/tree/master/test?ref_type=heads
which our new detector, https://github.com/wlynxg/chardet, fails on 31 of them:

```
testdata/uchardet/bg/windows-1251.txt: got ISO-8859-1, want Windows-1251 (confidence 0.99, language Bulgarian
testdata/uchardet/da/iso-8859-15.txt: got ISO-8859-1, want ISO-8859-15 (confidence 0.73, language 
testdata/uchardet/et/iso-8859-13.txt: got ISO-8859-1, want ISO-8859-13 (confidence 0.73, language 
testdata/uchardet/et/windows-1257.txt: got ISO-8859-1, want Windows-1257 (confidence 0.73, language 
testdata/uchardet/fi/iso-8859-1.txt: got MacRoman, want ISO-8859-1 (confidence 0.7159344894026975, language 
testdata/uchardet/he/iso-8859-8.txt: got ISO-8859-1, want ISO-8859-8 (confidence 0.99, language Hebrew
testdata/uchardet/he/windows-1255.txt: got ISO-8859-1, want Windows-1255 (confidence 0.9773686833361969, language Hebrew
testdata/uchardet/hr/iso-8859-13.txt: got ISO-8859-1, want ISO-8859-13 (confidence 0.73, language 
testdata/uchardet/hr/iso-8859-16.txt: got ISO-8859-1, want ISO-8859-16 (confidence 0.73, language 
testdata/uchardet/hr/iso-8859-2.txt: got ISO-8859-1, want ISO-8859-2 (confidence 0.73, language 
testdata/uchardet/hu/iso-8859-2.txt: got ISO-8859-1, want ISO-8859-2 (confidence 0.73, language 
testdata/uchardet/lt/iso-8859-10.txt: got ISO-8859-1, want ISO-8859-10 (confidence 0.73, language 
testdata/uchardet/lt/iso-8859-13.txt: got ISO-8859-1, want ISO-8859-13 (confidence 0.73, language 
testdata/uchardet/lt/iso-8859-4.txt: got ISO-8859-1, want ISO-8859-4 (confidence 0.73, language 
testdata/uchardet/lv/iso-8859-10.txt: got ISO-8859-1, want ISO-8859-10 (confidence 0.73, language 
testdata/uchardet/lv/iso-8859-13.txt: got ISO-8859-1, want ISO-8859-13 (confidence 0.73, language 
testdata/uchardet/lv/iso-8859-4.txt: got ISO-8859-1, want ISO-8859-4 (confidence 0.73, language 
testdata/uchardet/mt/iso-8859-3.txt: got ISO-8859-1, want ISO-8859-3 (confidence 0.73, language 
testdata/uchardet/no/iso-8859-15.txt: got ISO-8859-1, want ISO-8859-15 (confidence 0.73, language 
testdata/uchardet/pl/iso-8859-13.txt: got ISO-8859-1, want ISO-8859-13 (confidence 0.6365243902439024, language 
testdata/uchardet/pl/iso-8859-16.txt: got ISO-8859-1, want ISO-8859-16 (confidence 0.73, language 
testdata/uchardet/pl/iso-8859-2.txt: got ISO-8859-1, want ISO-8859-2 (confidence 0.73, language 
testdata/uchardet/ro/iso-8859-16.txt: got ISO-8859-1, want ISO-8859-16 (confidence 0.73, language 
testdata/uchardet/sk/iso-8859-2.txt: got ISO-8859-1, want ISO-8859-2 (confidence 0.6586976744186046, language 
testdata/uchardet/sl/iso-8859-16.txt: got ISO-8859-1, want ISO-8859-16 (confidence 0.73, language 
testdata/uchardet/sl/iso-8859-2.txt: got ISO-8859-1, want ISO-8859-2 (confidence 0.73, language 
testdata/uchardet/tr/iso-8859-3.txt: got ISO-8859-1, want ISO-8859-3 (confidence 0.73, language 
testdata/uchardet/tr/iso-8859-9.txt: got ISO-8859-1, want ISO-8859-9 (confidence 0.73, language 
```

That's over 20%, so I propose the following solution:

https://github.com/editorconfig-checker/editorconfig-checker/blob/f5540028e9ee248599578b95aac77998430d4ed4/README.md?plain=1#L378-L415

klaernie:
Reading the tests in the https://github.com/editorconfig/editorconfig-plugin-tests I'm pretty sure all the ISO-8859 variants are all grouped as `latin1`, or the understanding of `latin1` means pure ASCII.

So I think I would reduce the ISO8859 variants to all be treated as `latin1` - assuming that a user specifying `latin1` in `.editorconfig` intends to accept all variants of ISO8859, specifically the variant they use locally - no matter if it is ISO8859-1 or IS=8859-15.

The more important use case should be IMHO that we correctly identify UTF8, 16 and 32 in both their endiannesses and detect a BOM. This will be the more frequent use case for people wanting to ensure their codebase is up to a modern standard and not introducing files containing what I'd call legacy encoding.

rasa:
@klaernie Good feedback. I thought of that too. But what led me to think `latin1` means ISO-8859-1 is https://github.com/editorconfig/editorconfig/wiki/Character-Set-Support which links `latin1` to https://en.wikipedia.org/wiki/ISO/IEC_8859-1 . That page says it's "Latin alphabet no. 1"

But I think you're right. By default, we interpret `latin1` to mean "any text file that's not a UTF* encoding. But perhaps we add an option to interpret `latin1` as ISO-8869-1. Perhaps something like:
 ```json 
 { 
   ... 
   "Charsets": {
     "Latin1": ["ISO-8859-1"]
   }
   ... 
 } 
 ``` 
Thoughts?

pkg/encoding/encoding.go
```diff
@@ -93,92 +133,396 @@ var encodings = map[string]encoding.Encoding{
...
+// CharsetsEqual checks the two charset names for equality.
+func CharsetsEqual(charsetFound, charsetWanted string, bomEncoding string) bool {
+	charsetFound = normalizeCharsetName(charsetFound)
+	charsetWanted = normalizeCharsetName(charsetWanted)
+	if charsetWanted == "utf8bom" && bomEncoding != consts.UTF8 {
+		return false
+	}
+	if charsetFound == "utf8" && bomEncoding == consts.UTF8 {
+		charsetFound += "bom"
+	}
```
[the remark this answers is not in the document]
rasa: @ccoVeille Thanks for the feedback. I've attempted to simplify as you suggested. Lemme know if I hit the mark.

rasa:
Note, I had to add `"application/octet-stream",` to [validation.go](https://github.com/rasa/editorconfig-checker/blob/rs/support-charset-setting/pkg/validation/validation.go#L25)'s `textRegexes` list, for things to work as expected.

Now it's in sync (again) with [pkg/config/config.go](https://github.com/rasa/editorconfig-checker/blob/6f1c4bc95712e3b8f4045bafca7661520e9b59db/pkg/config/config.go#L72).

I'm not sure how things worked without this.

klaernie:
> @klaernie Good feedback. I thought of that too. But what led me to think `latin1` means ISO-8859-1 is https://github.com/editorconfig/editorconfig/wiki/Character-Set-Support which links `latin1` to https://en.wikipedia.org/wiki/ISO/IEC_8859-1 . That page says it's "Latin alphabet no. 1" https://www.iana.org/assignments/character-sets/character-sets.xhtml also says an alias for ISO-8859-1 is `latin1`.

I did not find the wiki page - that uncovers the hints I failed to notice.

So we should implement `latin1` as strictly ISO-8859-1, and not support any of the other ISO-8859-* variants for now, as the editorconfig itself does not list them as supported. According to spec we must ignore any value we do no implement and treat it as `charset = unset`, so there is no harm in not supporting the other variants.

This also underpins the sentiment I get from the editorconfig wiki - I read it as "use unicode as a first choice" - which I personally also find as the most reasonable.

I'll hopefully get to review the code later, but that might be not today.

rasa:
> So we should implement `latin1` as strictly ISO-8859-1, and not support any of the other ISO-8859-* variants for now, as the editorconfig itself does not list them as supported. According to spec we must ignore any value we do no implement and treat it as `charset = unset`, so there is no harm in not supporting the other variants.

I hear you. I initially thought so as well, but you convinced me to be more lenient given our false positives in identifying `iso-8859-1` files.

So, note that
https://github.com/editorconfig/editorconfig/wiki/Character-Set-Support#supported-character-sets says

> Other character sets could be specified outside of this set and they would be ignored if not understood by the editor.

Since we are not an editor, I think we can include support for other character sets. The spec says that `latin1` and the `utf*`s are what "all plugins should attempt to support at a minimum." It doesn't say we can't support others.

> This also underpins the sentiment I get from the editorconfig wiki - I read it as "use unicode as a first choice" - which I personally also find as the most reasonable.

Agreed. And I can see many people using editorconfig-checker to help them migrate their legacy files to Unicode.

> I'll hopefully get to review the code later, but that might be not today.

Take your time. I feel it's ready for review, but there are three small changes I am considering:
1. Add a test for [encoding.Supported](https://github.com/rasa/editorconfig-checker/blob/5bb70b9cadc7aae8b633080c76367e87192605bb/pkg/encoding/encoding.go#L553) which I somehow overlooked.
2. Use `sparse-checkout` in our encoding package's [Makefile](https://github.com/rasa/editorconfig-checker/blob/rs/support-charset-setting/pkg/encoding/testdata/Makefile), and
3. Use golang/text's [ascii encoder](https://github.com/golang/text/blob/4890c57b7721969ba8997aea0970c11004f1f5b7/encoding/ianaindex/ianaindex.go#L157) for ASCII file decoding.

pkg/encoding/encoding_test.go
```diff
@@ -1,185 +1,866 @@
...
-	for _, tt := range decodeTextTests {
-		filename := "testdata/" + tt.filename
-		rawFileContent, err := os.ReadFile(filename)
+const minConfidenceToFailTests = 1
+
+const minConfidenceToSetEncoding = 10
+
+const defaultConfidence = 1
+
+const testResultsJson = "test-results.json"
```
klaernie, line 28: do I understand this correctly as being a test snapshot file? If so, why not reuse snaps - there the order of the tests would not matter?

rasa: To be honest, I haven't worked with go-snaps before. Are you suggesting we should here?

Note: If `EDITORCONFIG_ADD_NEW_FILES=1` is set in the environment, the tests suite will scan for new files in `testdata` and add them to the `test-results.json`. Otherwise the suite runs only on the `testdata` files listed in the file.

I found the file very helping in my debugging, as I could run a `git diff` after a run to see if anything changed. That was a lot easier than scanning the log output for hundreds of files.

pkg/encoding/encoding.go
```diff
@@ -18,167 +20,606 @@ import (
...
+
+	// BinaryData contains the string to return if the data contains binary
+	// data (data that is not decodable, by any of the decoders that golang provides).
+	BinaryData = "binary"
+
+	// UnknownEncoding is returned if the encoding could not be determined.
+	UnknownEncoding = "unknown"
+
+	// See https://spec.editorconfig.org/#supported-pairs
+	// CharsetUnset defines the value allowing for file encoding.
```
rasa, line 49: We could use the constants at https://github.com/editorconfig/editorconfig-core-go/blob/7404f9a31780afcfa5bc3a62697476e720b39c5d/editorconfig.go#L36 instead.

pkg/encoding/encoding.go
```diff
@@ -18,167 +20,606 @@ import (
...
+// Per https://github.com/editorconfig-checker/editorconfig-checker/pull/457#issuecomment-2779587476
+func CharsetsMatch(charsetFound, charsetWanted string) bool {
+	charsetFound = normalizeCharsetName(charsetFound)
+	charsetWanted = normalizeCharsetName(charsetWanted)
+
+	if charsetFound == "utf8sig" {
+		charsetFound = "utf8bom"
+	}
+
+	if !supported(charsetFound) {
```
rasa, line 256: `supportedUTFEncoding()` would be a more apt name.

pkg/encoding/encoding.go
```diff
@@ -18,167 +20,606 @@ import (
...
+			encoding = detected
+			break
+		}
+
+		if encoding == consts.Ascii {
+			break
+		}
+
+		// We need to check for UTF16/32 encodings first, as
+		// UTF16/32 encoded first can be valid UTF8 files (surprisingly).
```
rasa, line 318: s/first/files/

rasa:
I think the test files at
https://github.com/arthenica/libiconv/tree/master/tests and
https://github.com/pa-0/dos2unix/tree/master/dos2unix/test
may be of higher quality, so I am thinking of adding them to our testdata.

pkg/encoding/encoding_test.go
klaernie: okay, so basically you implement the reverse behaviour of snaps. Snaps will add new snapshots, but never change existing snapshots unless `UPDATE_SNAPS` is set to a true value.

In this use case I would think using `snaps.MatchStandaloneJSON(t, someValue)` would be better.
https://github.com/gkampitakis/go-snaps?tab=readme-ov-file#matchjson

This would mean:
- always scan for testfiles
- during teardown call `snaps.MatchStandaloneJSON()` for each test. Although one might argue that matching the snapshot inline would be easier, but right now the test architecture touches the central `tests` slice multiple times, if I understand it correctly.

snaps itself will then generate test failures when a previously created snapshot is not matched.

Makefile
```diff
@@ -1,5 +1,6 @@
 ifeq ($(OS),Windows_NT)
 	STDERR=con
+	EXEEXT=.exe
```
klaernie, line 3: where did that come from?

klaernie:
> I think the test files at https://github.com/arthenica/libiconv/tree/master/tests and https://github.com/pa-0/dos2unix/tree/master/dos2unix/test may be of higher quality, so I am thinking of adding them to our testdata.

I wonder if there is an optimal set instead of collecting test files from everywhere. But currently I'm on neither side of the fence, so feel free to add them.

> If code review hasn't started, lemme know, and I'll set this to draft status, and add them, as well as the comments noted above.

There is no binary state of  code review, feel free to make changes as you see fit. We maintainers would be stupid to keep you from iterating towards the best solution - after all you're doing the hard work right now, and I'm very thankful for that!

---

Generally I'm a bit vary of the huge implementation of the test cases, since I still haven't wrapped my head around it fully. It seems fairly complicated, but probably isn't. But I think if you convert this from storing tests in a json file to matching snapshots it might become clearer.

Makefile
rasa: @klaernie Windows' exe's need an extension for Windows to execute them.

rasa:
The code I believe is production ready, but I'm converting to draft to revision the test framework, and explore using snapshots, as @klaernie suggested. Please be patient as I haven't worked with this tooling before. I think only using the highest quality test files, makes sense too.

rasa:
> Hey @rasa any progress on this?

@mstruebing No, sorry. I haven't looked into snapshots, which I think is a good idea. And reducing the number of test files probably makes sense too. I don't know when I'll get to these changes though. If someone has interest, I would welcome any help they are willing to offer.

rasa:
I rebased main onto this PR. I am going to change it to "Ready for review", as it is production ready, and so, it could be merged as is. But I understand if you want to wait until we add snapshot support. I don't have experience with snapshots, so there will be a learning curve.

mstruebing:
@klaernie @ccoVeille any opinion moving forward with this?

ccoVeille:
I feel like code is fine from my memories.

Things would be better than earlier with this PR.

It addresses issues in some languages, some errors may pop out with these changes, but I feel like it's better to merge and fix issues if something is reported than keep this code unmerged

mstruebing:
I've rebased this PR - there was a conflict in the `go.mod` file.
As soon as CI is green I will hit the merge button.
If something goes wrong, we can still fix it or in the worst case scenario, revert.

PR #464, opened 2025-04-28
fix: skip correct number of errors when consolidating errors

brainiac:
`ConsolidateErrors` is skipping over too many errors during consolidation, leaving some errors unreported.


theoludwig (changes requested):
Thank you for contributing.

Could you please add a test case to demonstrate the issue? Do you have a reproducible example?

brainiac:
Added a test case. without the fix, it will fail with the following:
```
=== RUN   TestConsolidateErrorCounts
    error_test.go:212: consolidation expectation          : [{2 wrong indentation type 3} {7 wrong indentation type 0} {9 wrong indentation type 2}]
    error_test.go:213: consolidation actual returned value: [{2 wrong indentation type 3} {11 wrong indentation type 0}]
    error_test.go:214: returned list of validation errors deviated from the expected set
--- FAIL: TestConsolidateErrorCounts (0.00s)

```

klaernie (approved):
Crazy that this pattern escaped me. But the fix is elegantly simple. Thanks a lot for the test case!

theoludwig (approved):

brainiac:
I updated the commit messages to [hopefully] make CI happy

ccoVeille (approved):

klaernie:
> I updated the commit messages to [hopefully] make CI happy
> 
> edit: Looks like I didn't get it right. I can check back on this later tonight - I don't want to flood this thread with approvals just to get the commit message right. Any guidance on this in the meantime would be appreciated.

It now only complains about the uppercase `Skip` and `Add`. You can also run the same test at commit time locally, if you `pre-commit install`.  Don't worry about the approvals - they are due to us are all in favour of your change, and hitting the "run CI" button is not that hard ;)

brainiac:
thank you :)

Issue #276, opened 2023-07-24
Add multi-thread support


bezirg:
This is a feature request.

For large projects, checking in multiple threads could parallelize the I/O and processing time, hopefully yielding better performance. 

Also, it should be worht to provide an option flag to control the number of threads used.

mstruebing:
I actually remember back when I was implementing this tool in its first iteration I tried to run it with go routines in multiple threads but the result was actually slower than a single thread.

Maybe that changed with all the changes in between, and I'm not entirely sure but I believe I've had an HDD back then, no SSD. So it may be worth trying and benchmarking it first before building a solid solution.

philosopherstonerush:
Hi @mstruebing, Any specific reason why you found it to be slower? Creating a lot of go routines might be a problem but I guess we could control it with a worker-pool. Given how each file is atomic, multi-threading could be simple to implement. I believe contention over the logger must be looked into as well (might be one of the reasons why it was slow) but we could overcome it if we were to have a separate goroutine that printed stuff into the console from a channel. 

I would be happy to take this issue up and see if it does speed up things.

mstruebing:
No I'm not entirely sure about the reasons why it was slower.

klaernie:
Technically, from having looked deeply into the Logger and Error structures, multithreading would need to feed the validation for an entire file back to the central logger. But users would be probably confused if the output would intermix violations from deeply different files in the structure.

But the effort is questionable - even big codebases are investigated within seconds - at least for what I consider big. Distributing this work across multiple CPUs would start making sense when the number of files crosses in the 100k.

As an example I took the worst case I have on my machine - a configuration repository containing 35k files, 207MB total:

    editorconfig-checker  8.11s user 2.54s system 154% cpu 6.886 total

Do you have any proper worst case that takes longer than this (with warm filesystem buffers)?

philosopherstonerush:
> Technically, from having looked deeply into the Logger and Error structures, multithreading would need to feed the validation for an entire file back to the central logger. But users would be probably confused if the output would intermix violations from deeply different files in the structure.

It is a problem that I thought of too but I don't think it would be a huge deal when it comes to this particular application, but maybe we could provide an option to sort it by common path and write to console. 

```
/b/c/a.txt
a1.txt
/b/c/a2.txt

// separate filenames from path (/b/c/, a.txt)
m := map[string]string[] {
  "/b/c/": [a.txt, a2.txt],
  "./": [a1.txt]
}
[...]

PR #488, opened 2025-08-27
perf: process files concurrently

ngotchac:
Fixes https://github.com/editorconfig-checker/editorconfig-checker/issues/276

This makes the whole run sub 2s on my "big"-ish project, instead of the previous ~30s, via a couple of tweaks:
1. Early skip a whole dir in the initial walking of the files, if it matches the exclude pattern
2. Process files concurrently
3. Pre-compile some regexps (it misses the indentation regexp, which is dynamic, but we could maybe pre-compile a couple of them (eg. 2 / 4 / 8)
4. Use a stream in `GetContentTypeBytes` so we don't need to read the whole file

It also adds a new `--cpuprofile` flag to write a CPU profile of the run to the given filepath

### DIFF
--- a/cmd/editorconfig-checker/main.go
+++ b/cmd/editorconfig-checker/main.go
@@ -6,6 +6,7 @@ import (
 	"flag"
 	"io/fs"
 	"os"
+	"runtime/pprof"
 	"strconv"
 	"strings"
 
@@ -42,15 +43,19 @@ const (
 )
 
 // these must be globals, since they are referenced by init(), parseArguments
-var configFilePath string
-var cmdlineExclude string
-var cmdlineConfig config.Config
-var writeConfigFile bool
+var (
+	configFilePath  string
+	cmdlineExclude  string
+	cmdlineConfig   config.Config
+	writeConfigFile bool
+	cpuprofile      string
+)
 
 func enableNoColor(string) error {
 	cmdlineConfig.NoColor = true
 	return nil
 }
+
 func disableNoColor(string) error {
 	cmdlineConfig.NoColor = false
 	return nil
@@ -78,6 +83,7 @@ func init() {
 	flag.BoolVar(&cmdlineConfig.Disable.Indentation, "disable-indentation", false, "disables the indentation check")
 	flag.BoolVar(&cmdlineConfig.Disable.IndentSize, "disable-indent-size", false, "disables only the indent-size check")
 	flag.BoolVar(&cmdlineConfig.Disable.MaxLineLength, "disable-max-line-length", false, "disables only the max-line-length check")
+	flag.StringVar(&cpuprofile, "cpuprofile", "", "write cpu profile to file")
 }
 
 // parse the arguments from os.Args
@@ -103,7 +109,7 @@ func parseArguments() {
 
 	flag.Parse()
 
-	var configPaths = []string{}
+	configPaths := []string{}
 	if configFilePath == "" {
 		configPaths = append(configPaths, defaultConfigFileNames[:]...)
 	} else {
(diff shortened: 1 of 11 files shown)


pkg/files/files.go
```diff
@@ -119,27 +135,12 @@ func GetFiles(config config.Config) ([]string, error) {
...
-				return err
-			}
-
-			if fi.Mode().IsRegular() {
-				filePaths = AddToFiles(filePaths, path, config)
-			}
-
-			return nil
-		})
+		return GetFilesFromDirectory(cwd, config)
```
ngotchac, line 138: Here, it previously didn't return when the `git` command returned an error, but I think it was a mistake?

theoludwig (approved):
Thanks! The changes seems reasonable to me, and it's a great performance improvement.

cmd/editorconfig-checker/main.go
```diff
@@ -155,7 +161,22 @@ func parseArguments() {
...
+		}
+		if err := pprof.StartCPUProfile(f); err != nil {
+			currentConfig.Logger.Error("Starting CPU profile: %v", err.Error())
+			exitProxy(exitCodeErrorOccurred)
+		}
+	}
+
 	config := *currentConfig
+	// force the exclude regexp to be compiled and cached
+	config.CachedExcludesAsRegexp()
```
ccoVeille, line 178: If I'm not wrong, this is the the first call to this method.

So it would cache.

But the regexp could fail, and right now it would panic

So here I would use this


```suggestion
	_, err = config.CachedExcludesAsRegexp()
	if err != nil {
		currentConfig.Logger.Error("Caching exclusions: %v", err.Error())
		exitProxy(exitCodeErrorOccurred)
	}
```

It comes with this

https://github.com/editorconfig-checker/editorconfig-checker/pull/488#discussion_r2311912998

pkg/config/config.go
```diff
@@ -305,14 +314,23 @@ func (c *Config) mergeDisabled(disabled DisabledChecks) {
...
 
+// CachedExcludesAsRegexp returns the excludes as a compiled regular expression
+// The regexp compilation is cached
+// Note: This is not thread-safe
+func (c *Config) CachedExcludesAsRegexp() *regexp.Regexp {
+	if c.excludeRegexp == nil {
+		c.excludeRegexp = regexp.MustCompile(c.GetExcludesAsRegularExpression())
+	}
+	return c.excludeRegexp
+}
```
ccoVeille, line 332: regexp.MustCompile could panic

So I would do this

```suggestion
func (c *Config) CachedExcludesAsRegexp() (*regexp.Regexp, error) {
	if c.excludeRegexp != nil {
		return c.excludeRegexp, nil
	}

	var err error
	c.excludeRegexp, err = regexp.MustCompile(c.GetExcludesAsRegularExpression())
	return c.excludeRegexp, err
}
```

This comes with

https://github.com/editorconfig-checker/editorconfig-checker/pull/488#discussion_r2311912094

pkg/config/config.go
```diff
@@ -305,14 +314,23 @@ func (c *Config) mergeDisabled(disabled DisabledChecks) {
 }
 
 // GetExcludesAsRegularExpression returns the excludes as a combined regular expression
-func (c Config) GetExcludesAsRegularExpression() string {
+func (c *Config) GetExcludesAsRegularExpression() string {
 	if c.IgnoreDefaults {
 		return strings.Join(c.Exclude, "|")
 	}
-
 	return strings.Join(append(c.Exclude, DefaultExcludes), "|")
```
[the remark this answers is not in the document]
ccoVeille: If there is a need (and I'm unsure there is), it's out of the scope of this PR.

So we can move forward.

PR #506, opened 2025-11-21
fix: support when charset is not set (fixes #505)

### LINKED ISSUE #505: Character encoding checks trigger on 'Wrong character encoding (_ instead of "")'
Great tool!

Since release 3.5.0 which seems to include file character encodings checks, I'm getting a few findings that are seemingly false positives:

```
> ec .

.editorconfig:
	Wrong character encoding ("Ascii" instead of "")
readme.md:
	Wrong character encoding ("UTF-8" instead of "")
```

my `.editorconfig`:

```
[*]
insert_final_newline = true

[*.{js,json,json5,ts,ps1,py,sql,tsql,yaml,yml}]
charset = utf-8
indent_style = space
trim_trailing_whitespace = true

[*.{md}]
charset = utf-8
indent_style = space

[*.py]
indent_size = 4

[*.{js,ts}]
indent_size = 2
```

The detected encodings seem correct (ascii and UTF-8), but it is compared against an undefined expected encoding.

rasa:
Fixes #505

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,12 +1,5 @@
 # Changelog
 
-## [3.5.1](https://github.com/editorconfig-checker/editorconfig-checker/compare/v3.5.0...v3.5.1) (2025-11-21)
-
-### Bug Fixes
-
-* support missing or empty charset config file setting ([#505](https://github.com/editorconfig-checker/editorconfig-checker/issues/505)) ([7214c58](https://github.com/editorconfig-checker/editorconfig-checker/commit/7214c58a91d158473eb3e2951e51ec788b877b90))
-
-
 ## [3.5.0](https://github.com/editorconfig-checker/editorconfig-checker/compare/v3.4.1...v3.5.0) (2025-11-18)
 
 


theoludwig (changes requested):

theoludwig (approved):
Thanks! That solves the problem.

Just pointing that out here, as it's related to charset:
I also have lot of issues with: `"UTF-8-SIG" instead of "utf-8"`, shouldn't we consider that it's basically the same thing, and should we be that strict? (that can probably be solved separately than this PR).
What do you think?

CHANGELOG.md
```diff
@@ -5,7 +5,7 @@
 
 ### Features
 
-* support chardet config file setting ([#457](https://github.com/editorconfig-checker/editorconfig-checker/issues/457)) ([f216ce5](https://github.com/editorconfig-checker/editorconfig-checker/commit/f216ce52132e5fbe1d8c7abdcccd6627b264e285))
+* support charset config file setting ([#457](https://github.com/editorconfig-checker/editorconfig-checker/issues/457)) ([f216ce5](https://github.com/editorconfig-checker/editorconfig-checker/commit/f216ce52132e5fbe1d8c7abdcccd6627b264e285))
```
ccoVeille: Wasn't it about chardet lib ? And so there was no typo ?

@theoludwig

rasa:
> `"UTF-8-SIG" instead of "utf-8"`, shouldn't we consider that it's basically the same thing, and should we be that strict?

I think that `utf-8` is for non-BOM files, since `utf-8-bom` is for BOM files. If we want to support both, perhaps we allow multiple charsets:
```
charset = utf-8,utf-8-bom
```
or maybe
```
charset = utf-8*
```

theoludwig:
@ccoVeille @mstruebing We should merge and release this PR rather quickly as it impacts lot of people as shown here: https://github.com/editorconfig-checker/action-editorconfig-checker/issues/236, #511, #509, #505

PR #556, opened 2026-04-10
fix: remove duplicated arch from ec-compat binary name

### LINKED ISSUE #208: Remove duplicated arch from binaries
The artifact archive file has the arch in the name already, which is duplicated with the binary inside and requires a rename to make the executable just `ec` on each platform. Please, consider removing the duplication! I am using a growing in popularity tool called [eget](https://github.com/zyedidia/eget), which safely installs binaries from GitHub and other places, but it cannot cleanly install `editorconfig-checker`.

mvanhorn:
## Summary

The ec-compat release archives (e.g. `ec-linux-amd64.tar.gz`) contain a binary named `bin/ec-linux-amd64`, duplicating the OS/architecture info already present in the archive filename. This prevents tools like [eget](https://github.com/zyedidia/eget) from cleanly installing the binary as just `ec`.

Changed the goreleaser ec-compat binary template from `bin/ec-{{.Os}}-{{.Arch}}` to `ec` so the archive contains a cleanly named `ec` binary.

The main `editorconfig-checker` build is unaffected. Only the legacy ec-compat archives change.

Fixes #208

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -36,7 +36,7 @@ builds:
   # this build is used to provide the legacy-style archives
   - id: "ec-compat"
     main: ./cmd/editorconfig-checker
-    binary: bin/ec-{{- .Os }}-{{ .Arch }}{{- if .Arm }}v{{ .Arm }}{{ end }}
+    binary: ec
     env:
       - CGO_ENABLED=0
     goos:


mstruebing:
This will need adjustments in all wrapping tools {javascript,python,php} as they use they use this binary name.
So in theory this is even a breaking change.

I'm not against it but someone needs to take care to adjust these properly.

mvanhorn:
Good point about the wrapper tools. Happy to open PRs for the JS, Python, and PHP wrappers to update the binary name references. Should I do that in parallel so they can all land together, or would you prefer to handle the coordination?

theoludwig:
I feel like the breaking change is not worth it yet.
And if ever do one, we might do "everything at once" meaning we also fix #371 at the same time.

Can you please elaborate further on this:
> This prevents tools like [eget](https://github.com/zyedidia/eget?rgh-link-date=2026-04-10T08%3A30%3A46Z) from cleanly installing the binary as just ec.

Why it is not possible with the current archives.

mvanhorn:
Fair point about batching breaking changes. For context on the eget use case: eget extracts a binary from the archive and renames it based on the tool name. When the binary inside is `ec-linux-amd64` instead of just `ec`, eget can't auto-detect the right binary name and users have to pass `--file` manually. Same issue affects aqua and other binary installers that expect a predictable binary name inside archives.

That said, bundling this with #371 makes sense if you're planning a breaking release anyway. Happy to wait and contribute both changes together when the time comes. Should I close this for now and reopen when #371 is ready?

klaernie:
@mvanhorn does the tooling work using the new name `editorconfig-checker`? After all the legacy archives are just there until all wrappers and users have migrated to the new name.

mvanhorn:
@klaernie - yes, the new `editorconfig-checker-*` archives already work cleanly. Verified against v3.6.1:

```
$ tar -tzf editorconfig-checker-linux-amd64.tar.gz
CHANGELOG.md
LICENSE
README.md
editorconfig-checker
```

The binary sits at archive root with the clean name. `eget editorconfig-checker/editorconfig-checker` should resolve it without `--file`. The duplicated-arch issue only affects the legacy `ec-*` archives (those still have `bin/ec-linux-amd64`).

So this PR is genuinely scoped to ec-compat only. Given @theoludwig's earlier note about batching breaking changes with #371, the practical migration path for users is already "use the new archive name", and this PR is just a nice-to-have for stragglers still on the legacy `ec-compat`. Happy to keep it on hold for the #371 wave, or close it - your call.

klaernie:
Okay, so I would decide to close this, and assume we do not touch the legacy archives until they get removed eventually.

mvanhorn:
Sounds right to me, @klaernie. The legacy `ec-compat-darwin-darwin-arm64` name reaches a tiny audience and the migration cost would be visible to anyone scripting against the old names. Holding off until the legacy archives age out is the more conservative call. Thanks for the back-and-forth.

PR #557, opened 2026-04-16
fix(cli): auto-enable no-color when output format is github-actions

tmchow:
## Summary

Force `no-color` when the effective output format is `github-actions`, regardless of whether it was set via the `-format` flag or the project config file. This is the fix agreed on in the issue thread ("if format is 'github-actions', automatically set no-color").

## Why

GitHub Actions only recognizes the `::error file=...,line=...::msg` annotation syntax when the output is plain text. When ecc emits colorized output under `-format github-actions`, the ANSI escape sequences break the annotation parser and the errors never make it onto the pull-request files tab.

The maintainer spelled this out in [#537](https://github.com/editorconfig-checker/editorconfig-checker/issues/537):

> Should that mean, that we should automatically, if format is "github-actions", set no-color?

And the reporter confirmed that `editorconfig-checker -no-color -format github-actions` already works — the fix is to stop making users specify both.

## Changes

Single insertion in `cmd/editorconfig-checker/main.go`, right before `currentConfig.Merge(cmdlineConfig)`:

- Compute the effective output format using the same rule `Config.Merge` uses (a valid CLI format overrides the config-file format). If it resolves to `github-actions`, flip `cmdlineConfig.NoColor = true` so the merge propagates it into the logger.
- The check runs **before** merge so `c.Logger.Configure(...)` picks up `NoColor`; if it ran after merge the logger would already be configured and the CI output would still be colorized.
- `--format default` over a config-file `github-actions` correctly stays colored, because `cmdlineConfig.Format.IsValid()` takes the CLI value as authoritative.

## Testing

- Added two snapshot cases to `TestMainColorSupport`:
  - `format-github-actions-no-color` — `--format github-actions` alone produces uncolored `::error file=... ::` annotations.
  - `format-github-actions-color-arg-ignored` — `--color --format github-actions` still produces uncolored annotations; the `--color` override is intentionally ignored here because colorized annotations break the CI use case.
- `go test ./...` passes across all packages.
- Manually ran `editorconfig-checker --format github-actions testdata/trailing-whitespace.txt` and confirmed the output is plain `::error file=...,line=1::Trailing whitespace`.

Fixes #537

This contribution was developed with AI assistance (Codex).

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)

### DIFF
--- a/cmd/editorconfig-checker/__snapshots__/main_test.snap
+++ b/cmd/editorconfig-checker/__snapshots__/main_test.snap
@@ -118,9 +118,9 @@ testdata/trailing-whitespace.txt:
 
 ---
 
-[TestMainColorSupport/format-github-actions-color-arg-ignored - 1]
-::error file=testdata/trailing-whitespace.txt,line=1::Trailing whitespace
-
-1 errors found
+[TestMainColorSupport/format-github-actions-color-override - 1]
+[31;1m::error file=testdata/trailing-whitespace.txt,line=1::Trailing whitespace[33;0m
+[31;1m
+1 errors found[33;0m
 
 ---
--- a/cmd/editorconfig-checker/__snapshots__/main_test.snap
+++ b/cmd/editorconfig-checker/__snapshots__/main_test.snap
@@ -110,3 +110,17 @@ testdata/trailing-whitespace.txt:
 1 errors found
 
 ---
+
+[TestMainColorSupport/format-github-actions-no-color - 1]
+::error file=testdata/trailing-whitespace.txt,line=1::Trailing whitespace
+
+1 errors found
+
+---
+
+[TestMainColorSupport/format-github-actions-color-arg-ignored - 1]
+::error file=testdata/trailing-whitespace.txt,line=1::Trailing whitespace
+
+1 errors found
+
+---
--- a/cmd/editorconfig-checker/main.go
+++ b/cmd/editorconfig-checker/main.go
@@ -155,6 +155,20 @@ func parseArguments() {
 		}
 	}
 
+	// GitHub Actions annotations do not parse ANSI color codes; they render
+	// as literal escape sequences and break the annotation format. Force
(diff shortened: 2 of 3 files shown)


theoludwig:
Thanks for the PR!

While, it's a good idea to default to no color if format GitHub is used.

> the `--color` override is intentionally ignored here because colorized annotations break the CI use case.

I still feel like that if a user set --color either as a CLI flag or in the JSON config, it means it should use it.

The defaults can change, but if a user ask for a feature, we should not block it, because we're trying to be smart.

That's the very definition of an override.
User is responsible enough, if use the tool outside the defaults.

---

I will not block the merge of the PR because of this, if others maintainers approves the changes, as it's true that it's unlikely that someone want to set color with format GitHub.

However would still prefer that CLI or config .json always override take predecences over the defaults.

While having smart default is good. :+1:

tmchow:
Good point @theoludwig. Updated so an explicit `--color` or `--no-color` flag always takes precedence over the format-based default. The auto-disable now only triggers when the user hasn't passed either flag.

The sentinel resets in `parseArguments()` alongside the other globals so it's safe for multi-call test scenarios.

All 93 tests pass. The snapshot for the override case now shows colored github-actions output, confirming the flag is honored.

mstruebing:
I agree with @theoludwig here without looking at the full PR.
We can try to be smart with defaults but an explicit set should be respected.

theoludwig (approved):

mstruebing (approved):

hashar:
9f4014ce0944f601472e5cbfaec31f711890c780 includes the 6.2MBytes binary file `/editorconfig-checker`. I am dropping it with #572.

I did not bother analyzing the binary (and you probably don't want to run it), I am assuming good faith and expecting it was a mistake. Ideally the binary should be removed from the repository but I don't think allows banning a file.

PR #558, opened 2026-04-21
feat(files): expand glob patterns in passed-file args (#190)

### LINKED ISSUE #190: PassedFiles glob as inline parameter
I didn't find how to pass `PassedFiles` as inline parameters in the README, so went to source code and found this [function](https://github.com/editorconfig-checker/editorconfig-checker/blob/d243b14ec3643aad59735d94ab24b248b62c7c2c/pkg/files/files.go#L72). It would be great to add handling between the loader from the config and the output of the `git ls-files`. Unfortunately, I don't write programming on Go yet, so I can't do it myself.

**Why is it necessary**
For example, when we use monorepo, sometimes we no need to check the entire project, but only part of it. Ye, we can use configs and pass them on, but it's not very convenient.

mvanhorn:
## Summary

Expands shell-style glob patterns passed via `PassedFiles` (command-line args), so invocations like `editorconfig-checker 'src/*.go'` work even when the shell doesn't expand the pattern for us.

Closes #190

## Why

`PassedFiles` currently handles two cases - a file path or a directory path - and falls through to `AddToFiles` for anything else. A quoted glob pattern was passed straight to `os.Stat`, which failed, and then `GetContentType` surfaced the same path as a missing file. Issue #190 asked for inline glob support so users can narrow down a monorepo scan without invoking the shell.

## Changes

- `pkg/files/files.go`: new `hasGlobMeta` + `resolvePassedFile` helpers. In `GetFiles`, each passed file is first resolved to one or more concrete paths: existing paths pass through unchanged, glob-shaped paths go through `filepath.Glob`, and non-glob missing paths are returned as-is so the existing error path still fires.
- `pkg/files/files_test.go`: three new tests cover a matching glob (`./../../*.md`), a non-glob missing path (no expansion, existing behavior), and a no-match glob (no expansion).
- `README.md`: `PassedFiles` row updated to mention glob patterns and the no-match fallthrough.

## Testing

```
go test ./...                       # all packages green
go test ./pkg/files/ -v -run Glob   # new tests green
go vet ./...                        # clean
gofmt -l pkg/files/                 # clean
```

## Scope

`path/filepath.Glob` supports `*`, `?`, and `[...]`. It does not support `**` - supporting that would need a dependency (e.g. `doublestar`), which felt out of scope for this GFI. Users who need recursive matching can still pass a directory path and rely on the existing recursive walk.

### DIFF
--- a/README.md
+++ b/README.md
@@ -271,7 +271,7 @@ A sample configuration file can look like this and will be used from your curren
 | `NoColor` | bool | `false` | Disable colored output |
 | `Exclude` | string[] | `[]` | Regular expressions for files to exclude from checking |
 | `AllowedContentTypes` | string[] | `[]` | Additional content types to check (added to the defaults listed below) |
-| `PassedFiles` | string[] | `[]` | Explicit list of files or directories to check. When set, only these paths are checked instead of auto-discovering files from the working directory or git |
+| `PassedFiles` | string[] | `[]` | Explicit list of files, directories, or shell-style glob patterns (e.g. `src/*.go`) to check. When set, only these paths are checked instead of auto-discovering files from the working directory or git. Glob patterns that don't match any file are left as-is so a subsequent content-type check surfaces the missing path |
 | `Version` | string | `""` | When set, the tool verifies this value matches the binary version and exits with an error if they differ. Useful for pinning a specific version in CI |
 | `Disable` | object | | Selectively disable individual checks (see below) |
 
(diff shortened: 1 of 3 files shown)


theoludwig (approved):
Thanks

mvanhorn:
Thanks @theoludwig. Pushed d27a024 with test coverage for the uncovered branches codecov flagged:

- `hasGlobMeta`: star/question/bracket plus the backslash-escape arm (Unix-only, since `filepath.Match` treats `\` as a path separator on Windows)
- `resolvePassedFile`: invalid-pattern error path
- `GetFiles`: invalid pattern propagates, and a glob that matches a directory is walked via `GetFilesFromDirectory`

`pkg/files` coverage 79.1% → 82.2%. `hasGlobMeta` and `resolvePassedFile` now at 100%.

PR #560, opened 2026-04-30
docs(help): add synopsis line documenting positional FILE args (#455)

### LINKED ISSUE #455: support passing files that should be checked
Right now editorconfig-checker doesn't allow passing files that should be checked as arguments, making it non-compliant with the treefmt formatter spec: https://treefmt.com/formatters-spec/#rules

Would be cool if it would optionally be able to pass files that should be checked, and only check these files in that case.

mvanhorn:
Closes #455.

## What changed

`--help` output previously jumped straight from `USAGE:` into the flag default dump, with no synopsis line and no mention that positional `FILE` args are accepted. Added a synopsis and a one-paragraph explanation before the OPTIONS section:

\`\`\`
USAGE:
  editorconfig-checker [OPTIONS] [FILE...]

With no FILE arguments, all files tracked by git are checked. When one or
more FILE arguments are given, only those files are checked (the configured
exclude patterns still apply).

OPTIONS:
  ...
\`\`\`

## Why this matters

The behavior already exists (handled in pkg/files/files.go). The issue reporter (@katexochen) said "Maybe we cav add a line to the help output to document the synopsis," and @klaernie agreed: "I was thinking the same, but didn't manage to publish a PR yet. So let's keep the issue as a reminder." This PR is exactly that note acted on — no functional change, just making the supported invocation discoverable from `--help`.

## Verification

- `go test ./...` — all passing
- `editorconfig-checker --help` now shows the synopsis above the flag list

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,25 +139,36 @@ brew install editorconfig-checker
   paru -S editorconfig-checker-git
   ```
 
-### Language Specific Wrappers:
+### Language Specific Wrappers
 
 We do have a couple of language specific wrappers available that you can install with your favorites language package manager.
 See their respective documentation for installation and usage instructions.
 
-* JavaScript/TypeScript: [npm package](https://www.npmjs.com/package/editorconfig-checker)
-* Python: [PyPI package](https://pypi.org/project/editorconfig-checker/)
-* PHP: [Composer package](https://packagist.org/packages/editorconfig-checker/editorconfig-checker)
+- JavaScript/TypeScript: [npm package](https://www.npmjs.com/package/editorconfig-checker)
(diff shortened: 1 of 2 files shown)


mstruebing (approved):
Thank you

theoludwig:
Can we update the usage section of the README as well?

https://github.com/editorconfig-checker/editorconfig-checker#usage

theoludwig:
@mvanhorn Can you please update your PR?

That way, I believe after, we'll release v3.7.0. #553 

**EDIT:** I did it.

mvanhorn:
@theoludwig appreciate the merge. Adding the synopsis line above the flag dump makes `--help` actually useful for new users instead of just a flag inventory.

PR #568, opened 2026-06-11
fix: anchor node_modules and target default excludes at path start

### LINKED ISSUE #567: Default exclude /node_modules/ does not match a top-level node_modules in directory-walk mode
## Summary

The default exclude pattern for NodeJS is `"/node_modules/"` (with a leading slash):

https://github.com/editorconfig-checker/editorconfig-checker/blob/main/pkg/config/config.go#L47

Exclude patterns are matched against paths relative to the repository root. A top-level `node_modules/` directory produces paths like `node_modules/@babel/code-frame/package.json`, which have no slash before `node_modules`, so the default exclude never matches them. Nested ones (`packages/*/node_modules/...`) are correctly excluded.

The same applies to other defaults with a leading slash, e.g. `"/target/"` (Cargo) for a top-level `target/` directory.

## Why this usually goes unnoticed

In a git repository the file list comes from `git ls-files --cached --others --exclude-standard`, where ignored directories never appear, so the exclude pattern is irrelevant. The bug only shows in the directory-walk fallback, i.e. when `git ls-files` fails (not a git repo, or git errors out). We hit it in GitLab CI: the job image runs as a different uid than the cloned repo, `git ls-files` failed with `fatal: detected dubious ownership`, the checker silently fell back to walking the filesystem, and then checked thousands of files in the top-level `node_modules/`.

## Reproduction

```sh
mkdir repro && cd repro            # intentionally NOT a git repo, to force the directory walk
printf 'root = true\n\n[*]\ninsert_final_newline = true\n' > .editorconfig
mkdir -p node_modules/foo packages/a/node_modules
printf 'x' > node_modules/foo/bad.js
printf 'x' > packages/a/node_modules/bad.js
editorconfig-checker
```

Result with v3.7.0:

```
node_modules/foo/bad.js:
	Wrong line endings or no final newline

1 errors found
```

`packages/a/node_modules/bad.js` is correctly excluded, `node_modules/foo/bad.js` is not.

## Suggested fix

Anchor the patterns so they also match at the start of a relative path, e.g.:

```go
"(^|/)node_modules/",
"(^|/)target/",
```

Happy to send a PR if that approach sounds right.

As an aside, the silent fallback from `git ls-files` to a full directory walk made this hard to diagnose; a warning on stderr when the fallback is taken would have helped.

zckrs:
## Problem

The default excludes `/node_modules/` and `/target/` start with a literal slash, but exclude patterns are matched against paths relative to the repository root. A top-level `node_modules/` directory produces paths like `node_modules/foo/index.js` with no slash before `node_modules`, so it was never excluded. Nested directories (`packages/a/node_modules/...`) were correctly excluded.

This only shows in directory-walk mode, since `git ls-files` already omits ignored files, which is why it went unnoticed. See #567 for the full analysis and reproduction.

## Fix

Anchor both patterns as `(^|/)node_modules/` and `(^|/)target/` so they match at the start of a relative path as well as after a slash, without matching names that merely end with the directory name (e.g. `my_node_modules/`).

- `pkg/config/config.go`: anchor the two patterns
- `pkg/config/config_test.go`: update the expected joined regex, add `TestDefaultExcludesAnchoring` covering top-level and nested paths plus the `my_node_modules`/`not_target` non-matches
- `README.md`: update the documented default excludes list

Verified against the reproduction from #567: a top-level `node_modules/foo/bad.js` without final newline is now excluded in walk mode (exit 0), and `go test ./pkg/...` passes.

Fixes #567

### DIFF
--- a/README.md
+++ b/README.md
@@ -394,7 +394,7 @@ Unless you choose to [ignore them](#ignoring-default-excludes), these paths are 
 // package manager, generated, & lock files
 // Cargo (Rust)
 "Cargo\\.lock$",
-"/target/",
+"(^|/)target/",
 // Composer (PHP)
 "composer\\.lock$",
 // RubyGems (Ruby)
@@ -410,7 +410,7 @@ Unless you choose to [ignore them](#ignoring-default-excludes), these paths are 
 "\\.mvn/wrapper/MavenWrapperDownloader\\.java$",
 "mvnw(\\.cmd)?$",
 // NodeJS
-"/node_modules/",
+"(^|/)node_modules/",
 // npm (NodeJS)
 "npm-shrinkwrap\\.json$",
 "package-lock\\.json$",
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -28,7 +28,7 @@ var defaultExcludes = []string{
 	// package manager, generated, & lock files
 	// Cargo (Rust)
 	"Cargo\\.lock$",
-	"/target/",
+	"(^|/)target/",
 	// Composer (PHP)
 	"composer\\.lock$",
 	// RubyGems (Ruby)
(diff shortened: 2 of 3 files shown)


klaernie (approved):
Color me impressed, very thorough implementation!

theoludwig (approved):

mstruebing (commented):
Thanks 🚀

PR #569, opened 2026-06-13
chore: add JSON schema for editorconfig-checker configuration

### LINKED ISSUE #194: Add Configuration File Schema into SchemaStore

3w36zj6:
close #194

This PR adds a JSON Schema for the editorconfig-checker JSON configuration file. This enables JSON Schema-aware editors and tools to validate `.editorconfig-checker.json` and provide completion for its configuration properties.

If this is accepted, it would be even more convenient to distribute versioned schemas from <https://editorconfig-checker.github.io/>, or to publish the schema through the [JSON Schema Store](https://www.schemastore.org/).

### DIFF
--- a/.editorconfig-checker.json
+++ b/.editorconfig-checker.json
@@ -1,10 +1,11 @@
 {
+    "$schema": "./.editorconfig-checker.schema.json",
     "Verbose": false,
     "Debug": false,
     "IgnoreDefaults": false,
     "SpacesAfterTabs": false,
     "NoColor": false,
-    "exclude": [
+    "Exclude": [
         "testfiles",
         "testdata"
     ],
--- a/.editorconfig-checker.schema.json
+++ b/.editorconfig-checker.schema.json
@@ -1,0 +1,135 @@
+{
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "title": "editorconfig-checker configuration",
+    "description": "Configuration for editorconfig-checker",
+    "type": "object",
+    "additionalProperties": false,
+    "properties": {
+        "$schema": {
+            "type": "string",
+            "description": "A field for the JSON schema specification."
+        },
+        "Version": {
+            "type": "string",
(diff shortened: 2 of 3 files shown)


theoludwig (approved):
LGTM, thanks!

.editorconfig-checker.json
```diff
@@ -1,10 +1,11 @@
 {
+    "$schema": "./.editorconfig-checker.schema.json",
     "Verbose": false,
     "Debug": false,
     "IgnoreDefaults": false,
     "SpacesAfterTabs": false,
     "NoColor": false,
-    "exclude": [
+    "Exclude": [
```
mstruebing, line 8: Will the change in casing cause any trouble?

klaernie: IMHO yes - although I like it for the consistency. So I would not include this lightly - since there are thousands of config files out there, which might use the lowercase version.

theoludwig: But don't both work? All existing configurations that still use the lowercase version will continue to work, right?

3w36zj6: JSON configuration file properties are derived from the `Config` struct and are consistently written in PascalCase. In my understanding, the fact that only `exclude` is written in lowercase in the documentation has no meaning beyond being a leftover from `.ecrc`.

https://github.com/editorconfig-checker/editorconfig-checker/blob/9e1c08eddf6ee63a89052e476555a8d0bd4a0339/pkg/config/config.go#L124-L152

However, this is a matter of the schema definition. Go's JSON decoder ignores case when matching field names, so this does not affect existing users. Only new users who rely on the schema will need to review their configuration.

mstruebing (approved):

PR #572, opened 2026-06-18
chore: remove binary editorconfig-checker at root of repo

hashar:
In 9f4014ce0 (#557) a 6.2MBytes editorconfig-checker binary has been introduced to the repository. Remove it and add the file to `.gitignore` to prevent it from being added later.

PR #575, opened 2026-06-24
fix: honor -no-color for .ecrc deprecation warning

### LINKED ISSUE #574: -no-color does not work with codeclimate format
Running something like `editorconfig-checker -no-color -format codeclimate` in a directory with a deprecated `.ecrc` file,
gives a colored warning in stdout, like this:
```
[33;1mThe default configuration file name `.ecrc` is deprecated. Use `.editorconfig-checker.json` instead. You can simply rename it[33;0m
```

Found with editorconfig-checker 3.7.0 inside an Alpine Linux 3.23 container.

theoludwig:
Fixes #574

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,5 @@
 /bin/
-editorconfig-checker
+./editorconfig-checker
 editorconfig-checker.test
 coverage.txt
 result-bin
--- a/cmd/editorconfig-checker/main_test.go
+++ b/cmd/editorconfig-checker/main_test.go
@@ -125,6 +125,22 @@ func TestMainEcrcDeprecationWarningHonorsNoColor(t *testing.T) {
 	}
 }
 
+func TestMainEcrcDeprecationWarningHonorsNoColorEnvVar(t *testing.T) {
+	t.Setenv("NO_COLOR", "1")
+	output, lastSeenCode := runWithArguments(t, "--config", "testdata/.ecrc")
+	if lastSeenCode != exitCodeNormal {
+		t.Errorf("main exited with return code %d, but we expected %d", lastSeenCode, exitCodeNormal)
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -1,5 +1,5 @@
 /bin/
-editorconfig-checker
```
theoludwig, line 2: This was added in #572 

However, the source code is in  `‎cmd/editorconfig-checker`, so there was a warning:
> git add -A -- cmd/editorconfig-checker/main_test.go
The following paths are ignored by one of your .gitignore files:
cmd/editorconfig-checker
hint: Use -f if you really want to add them.
hint: Disable this message with "git config set advice.addIgnoredFile false"

So instead, we specifcy `./` to explictly ignore the binary that could be at the root of the repo, but still allow `cmd/editorconfig-checker` changes.

cmd/editorconfig-checker/main_test.go
```diff
@@ -110,6 +110,21 @@ func TestMainWithEcrc(t *testing.T) {
 	}
 }
 
+func TestMainEcrcDeprecationWarningHonorsNoColor(t *testing.T) {
```
ccoVeille, line 113: Could you also check the bevahior of the code with env var NOCOLOR

I feel like your fix also implies this to be fixed

theoludwig: Added a unit test for the NOCOLOR env var. :+1:

theoludwig: (tests are quite redundant and repetitive, it might be worth at one point to refactor and use a proper testing framework, but that's unrelated to this PR)

ccoVeille (approved):

mstruebing (approved):

Issue #390, opened 2024-10-13
Remove deprecation warnings and make the necessary BREAKING CHANGES


theoludwig:
Tracking issue to not forget to remove the added deprecation warnings in v3.1.0 when we work for v4.0.0.
- #389
- #386

klaernie:
> During the [discussion of the default excludes](https://github.com/editorconfig-checker/editorconfig-checker/pull/462#issuecomment-2840244252) we discussed whether or not folders should have a `/` prefix to better differentiate between files and folders.
> 
> This change was not included in the PR because it is perceived as a breaking feature. Should this be potentially be tackled with a 4.0 release?

If you like, feel free to open a PR

PR #598, opened 2026-08-15
fix!: do not accept latin1 files as utf-8

### LINKED ISSUE #597: charset=utf-8 does not warn about ISO-8859-1 file
`editorconfig-checker` does not complain that a file is not encoded with UTF-8 even though `charset = utf-8` is set.

.editorconfig:
```.editorconfig
root = true

[*]
charset = utf-8
```

Create a text file encoded with ISO-8859-1:
```sh
printf 'some\xa0text' > test.txt
```
Hex a0 is a non-breaking space in ISO-8859-1. As far as I can tell, this file is not valid UTF-8.

Check that the contents are correct with `xxd test.txt`:
```
00000000: 736f 6d65 a074 6578 74                   some.text
```

`file --mime test.txt`:
```
test.txt: text/plain; charset=iso-8859-1
```

If I change `charset = utf-8` to `charset = ascii`, it complains as expected:
```
test.txt:
	Wrong character encoding ("ISO-8859-1" instead of "ascii")

1 errors found
```

Eljees:
Fixes #597.

`CharsetsMatch` treats `latin1`, `iso88591` and `ascii` as all satisfying `charset = utf-8`:

    latin1Encodings = []string{CharsetLatin1, "ascii", "iso88591"}
    // latin1 (iso88591), and ascii files are utf8 files, too.
    if charsetWanted == "utf8" && slices.Contains(latin1Encodings, charsetFound) {
        return true
    }

For `ascii` that is sound — ascii is a strict subset of utf-8. For latin1 it is not. Every byte value 0x80–0xFF is a legal latin1 character, and most of them are not legal utf-8: a lone `0xA0` or `0xFF` will make Go, Python or a browser refuse to decode the file. So a file that nothing downstream can read as utf-8 passed the check silently. This drops latin1 and iso88591 from the exemption and keeps ascii.

### Before / after on the reporter's case

`.editorconfig` with `charset = utf-8`, three files — one latin1, one ascii, one utf-8:

| | before | after |
| --- | --- | --- |
| `latin1.txt` (`caf\xe9 na\xefve`) | not reported | `Wrong character encoding ("ISO-8859-1" instead of "utf-8")` |
| `ascii.txt` | not reported | not reported |
| `utf8.txt` (`caf\xc3\xa9`) | not reported | not reported |
| exit code | 0 | 1 |

### What this costs on real code

The risk with a change like this is new warnings on files people are happy with, so I measured it rather than guessing. I built the binary at `123ba31` and with this patch, pointed both at a corpus of **11 937 files** (this repository, three unrelated Python projects, and the full Go standard library source tree) under a `.editorconfig` forcing `charset = utf-8`, and diffed the two outputs:

    before: 524 errors found
    after:  525 errors found

**One** new warning, on:

    gosrc/cmd/go/testdata/script/get_panic_issue75251.txt:
        Wrong character encoding ("ISO-8859-1" instead of "utf-8")

That file is a Go test fixture containing three raw `0xFF` bytes; its own text says `# contains 0xff`. Python refuses it with `'utf-8' codec can't decode byte 0xff in position 163`. So the single new warning is a file that genuinely is not utf-8 — the new behaviour is right about it.

For completeness: 895 files in that corpus are not valid utf-8, but the rest are object files, `.syso` resources and pprof profiles, which the checker classifies as `binary` and reports identically before and after.

### Tests

`go test ./...`: **10 packages ok, 0 FAIL** — before and after.

The new expectations are in `pkg/encoding/encoding_test.go` and `pkg/validation/validators/validators_test.go`. Note that `TestCharsetsMatch` already enumerated this pair and skipped it:

    if charset1 == "latin1" && charset2 == "utf-8" {
        continue
    }

Removing that `continue` is what turns the test red on the current code (`CharsetsMatch("latin1", "utf-8"): got true, want false`) and green with the fix, so the existing table now covers the case instead of stepping around it.

`gofmt -l` and `go vet` are clean on the changed files. The commit follows the conventional-commit format that `commitlint` enforces here.

AI-assisted (LLM used for drafting); the measurements above are mine.

### DIFF
--- a/pkg/encoding/encoding.go
+++ b/pkg/encoding/encoding.go
@@ -4,7 +4,6 @@ package encoding
 import (
 	"bytes"
 	"fmt"
-	"slices"
 	"strings"
 	"unicode/utf8"
 
@@ -220,12 +219,6 @@ var (
 	encodingToDecoderMap map[string]string
 
 	supportedUTFEncodingMap map[string]string
-
-	latin1Encodings = []string{
-		CharsetLatin1,
-		"ascii",
-		"iso88591",
-	}
 )
 
 func init() {
@@ -276,8 +269,10 @@ func CharsetsMatch(charsetFound, charsetWanted string) bool {
 		}
 	}
 
-	// latin1 (iso88591), and ascii files are utf8 files, too.
-	if charsetWanted == "utf8" && slices.Contains(latin1Encodings, charsetFound) {
+	// ascii is a strict subset of utf8, so ascii files are utf8 files, too.
+	// latin1 (iso88591) files are not: a lone 0xA0 byte, for instance, is
+	// valid latin1 but is not valid utf8.
+	if charsetWanted == "utf8" && charsetFound == "ascii" {
 		return true
 	}
 
--- a/pkg/encoding/encoding_test.go
+++ b/pkg/encoding/encoding_test.go
@@ -156,6 +156,7 @@ func TestCharsetsMatch(t *testing.T) {
 		{CharsetLatin1, "iso88591", false},
 		{CharsetLatin1, "iso8859-1", false},
 		{CharsetLatin1, "iso-8859-1", false},
+		{CharsetLatin1, "utf-8", false}, // issue #597: latin1 is not a subset of utf-8
 		{CharsetUTF8, "utf8", true},
 		{CharsetUTF8BOM, "utf8bom", true},
 		{CharsetUTF8BOM, "utf-8bom", true},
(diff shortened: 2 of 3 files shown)


klaernie (approved):
Looks good to me, however I would explicitly mention this as a breaking change in the release, that we stopped the incorrect behaviour of treating latin1 as valid utf-8.

theoludwig:
> Looks good to me, however I would explicitly mention this as a breaking change in the release, that we stopped the incorrect behaviour of treating latin1 as valid utf-8.

Indeed, from our rules, this change should be considered as a breaking change: https://github.com/editorconfig-checker/editorconfig-checker#semantic-versioning-policy

> Bug fixes, which result in editorconfig-checker reporting more linting errors, because the previous behavior was incorrect according to the [editorconfig specification](https://editorconfig.org/).

Is it finally time to release v4.0.0 and do #371, #390? :eyes:

klaernie:
> Is it finally time to release v4.0.0 and do #371, #390? 👀

I would say so. Time to look at all the wrappers for #371.

theoludwig (approved):

mstruebing (approved):

PR #599, opened 2026-08-15
fix: drop a path whose content type could not be read

VXNCXNX:
## What's broken

Passing a path that does not exist panics the process.

```
$ ec nofile_does_not_exist_xyz.txt
Could not get the ContentType of file: nofile_does_not_exist_xyz.txt
stat nofile_does_not_exist_xyz.txt: no such file or directory
panic: open nofile_does_not_exist_xyz.txt: no such file or directory
    .../pkg/validation.ValidateFileWithDefinition(...) validation.go:72
```

## The fix

`AddToFiles` logs the `GetContentType` error and then appends the path anyway, because the skip is guarded by `err == nil`. `os.ReadFile` in the validator then panics on it.

It now returns early when the content type could not be read. The `err == nil` in the next condition becomes redundant and is dropped.

Nothing downstream wants the unreadable path: all three call sites only collect the returned slice, and its only consumer is `ValidateFileWithDefinition`, which is the panic site.

After:

```
$ ec nofile_does_not_exist_xyz.txt
Could not get the ContentType of file: nofile_does_not_exist_xyz.txt
stat nofile_does_not_exist_xyz.txt: no such file or directory
$ echo $?
0
```

## Verification

`TestGetFilesGlobWithoutMetaKeepsPath` already described this behaviour in a comment, "logs and drops them (they have no content type)", but its loop only returned when it found the path, so it passed either way and asserted nothing. It now fails if the path survives, which is what its comment always meant.

Added a nonexistent-path case to `TestAddToFiles` too. With the change reverted both fail, reporting the path present in the result. `go test ./...` passes.

### DIFF
--- a/pkg/files/files.go
+++ b/pkg/files/files.go
@@ -65,10 +65,11 @@ func AddToFiles(filePaths []string, filePath string, config config.Config) []str
 	if err != nil {
 		config.Logger.Error("Could not get the ContentType of file: %s", filePath)
 		config.Logger.Error("%v", err.Error())
+		return filePaths
 	}
 	config.Logger.Debug("AddToFiles: detected ContentType %s on file %s", contentType, filePath)
 
-	if err == nil && !IsAllowedContentType(contentType, config) {
+	if !IsAllowedContentType(contentType, config) {
 		config.Logger.Verbose("Not adding %s to be checked, it does not have an allowed ContentType", filePath)
 		return filePaths
 	}
--- a/pkg/files/files.go
+++ b/pkg/files/files.go
@@ -100,12 +100,16 @@ func hasGlobMeta(path string) bool {
 // named it explicitly. Patterns are expanded via filepath.Glob; one that
 // matches nothing is returned unchanged, so an empty match is not an error.
 func resolvePassedFile(passedFile string) ([]string, error) {
-	if _, err := os.Stat(passedFile); err == nil {
+	_, statErr := os.Stat(passedFile)
+	if statErr == nil {
 		return []string{passedFile}, nil
 	}
 	if !hasGlobMeta(passedFile) {
-		// A plain path that is not there is a typo, not a pattern that matched nothing.
-		return nil, fmt.Errorf("%s: no such file or directory", passedFile)
+		// A plain path that is not usable is a user error, not a pattern that
+		// matched nothing. Wrap the stat error so a cause other than a missing
+		// file, a permission denied on the parent directory for instance, is
+		// reported as it is.
+		return nil, fmt.Errorf("resolving %s: %w", passedFile, statErr)
 	}
 	matches, err := filepath.Glob(passedFile)
 	if err != nil {
--- a/pkg/files/files.go
+++ b/pkg/files/files.go
@@ -95,16 +95,17 @@ func hasGlobMeta(path string) bool {
 }
 
 // resolvePassedFile expands a single --passed-file argument into one or more
-// concrete paths. Paths that exist on disk are returned unchanged. Paths that
-// don't exist but look like glob patterns are expanded via filepath.Glob; if
-// the pattern matches nothing the argument is returned unchanged so the caller
-// can surface a not-found error in the usual way.
+// concrete paths. Paths that exist on disk are returned unchanged. A path that
+// does not exist and holds no glob metacharacters is an error, since the user
+// named it explicitly. Patterns are expanded via filepath.Glob; one that
+// matches nothing is returned unchanged, so an empty match is not an error.
 func resolvePassedFile(passedFile string) ([]string, error) {
 	if _, err := os.Stat(passedFile); err == nil {
 		return []string{passedFile}, nil
 	}
 	if !hasGlobMeta(passedFile) {
-		return []string{passedFile}, nil
+		// A plain path that is not there is a typo, not a pattern that matched nothing.
+		return nil, fmt.Errorf("%s: no such file or directory", passedFile)
 	}
 	matches, err := filepath.Glob(passedFile)
 	if err != nil {
--- a/pkg/files/files_test.go
+++ b/pkg/files/files_test.go
@@ -235,21 +235,19 @@ func TestGetFilesGlobPattern(t *testing.T) {
 	}
 }
 
-func TestGetFilesGlobWithoutMetaKeepsPath(t *testing.T) {
+func TestGetFilesMissingPathWithoutMetaErrors(t *testing.T) {
 	cfg := config.NewConfig(nil)
 	cfg.PassedFiles = []string{"./does/not/exist.txt"}
 
-	files, err := GetFiles(*cfg)
-	if err != nil {
-		t.Fatalf("GetFiles(missing): unexpected error %s", err.Error())
+	// A path the user named explicitly, with no glob metacharacters, is a typo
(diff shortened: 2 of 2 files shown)


klaernie (approved):
Thanks!

theoludwig:
> ```sh
> $ echo $?
> 0
> ```

Why should we exit with a 0 status code, if a file doesn't exist? Shouldn't we exit with a non 0 status code, since it is an error/failure?

VXNCXNX:
You are right, and my PR description undersold it. It is worse than "not ideal": this is a regression in the exit code.

Measured just now, same argument, nonexistent file:

```
main (panic):   exit=2
this PR:        exit=0
```

So the panic at least failed loudly. My change turns a hard failure into a silent success, which is the worse of the two, and my `$ echo $?  0` in the description read as if that were fine. It was actually the bug you spotted.

The cause is that the exit code is driven only by `GetErrorCount(errors)` in `main.go`. Dropping the path in `AddToFiles` removes it from `filePaths` entirely, so nothing is validated, no error is recorded, and the run reports success.

I tried the obvious fix, returning a not-found error from `resolvePassedFile` for a path with no glob metacharacters. It gives what you are asking for:

```
$ ec nofile_does_not_exist_xyz.txt
nofile_does_not_exist_xyz.txt: no such file or directory
$ echo $?
1
```

but it fails `TestGetFilesGlobWithoutMetaKeepsPath` from #558, which asserts the opposite:

```
GetFiles(missing): unexpected error ./does/not/exist.txt: no such file or directory
```

That test is deliberate, and its name says so. So the "should a missing explicit path be an error" question is one you own, not one I should decide by quietly rewriting a test that encodes the current answer. I reverted that attempt rather than push it.

Which way do you want it?

1. **Missing explicit path is an error.** `resolvePassedFile` returns the error, exit 1, and `TestGetFilesGlobWithoutMetaKeepsPath` gets updated with a note about the intended change. Consistent with most linters, and it means `ec typo-in-filename.go` cannot pass CI.
2. **Keep it non-fatal, but still fail the run.** Path is dropped as in this PR, and a non-validation error counter feeds the exit code, so it exits non-zero without aborting a multi-file run.
3. **Split it.** Merge only the panic fix here since that part is uncontroversial, and handle the exit code separately.

I lean towards 1, since a file the user named explicitly and that is not there is a user error, and glob patterns matching nothing are already handled separately. But it changes behaviour for anyone passing paths that may not exist, so it is your call.

Happy to push whichever you pick.

*Disclosure: written with AI assistance (Claude Code). The exit codes above are from binaries I built from `main` and from this branch, and I ran the failing test myself.*

theoludwig:
I myself also use AI a lot, and do not write code by hands anymore since beginning of the year.

However please when talking to humans, talk to humans, do not reply with AI slop.
Talk naturally, and also have taste in what you produce, that's where human still matter, and can decide and supervise the AI, and do the real engineering work.

So in your opinion
> Which way do you want it?

What do you think is the best option?

VXNCXNX:
Option 1: a file you named explicitly that is not there is a user error, so it should exit 1, and I would update the #558 test since it was written for globs, not for typos in a filename. Say go and I will push it.

klaernie:
I would also expect that we error out, when the user specifies a file that does not exist.

VXNCXNX:
Pushed. A path with no glob metacharacters that is not on disk now errors, so `editorconfig-checker ./does/not/exist.txt` exits 1 instead of 0.

A glob that matches nothing is still not an error, since that is a pattern doing its job rather than a typo. I updated the #558 test accordingly, it was written for the glob case.

pkg/files/files.go
```diff
@@ -94,16 +95,17 @@ func hasGlobMeta(path string) bool {
...
+// named it explicitly. Patterns are expanded via filepath.Glob; one that
+// matches nothing is returned unchanged, so an empty match is not an error.
 func resolvePassedFile(passedFile string) ([]string, error) {
 	if _, err := os.Stat(passedFile); err == nil {
 		return []string{passedFile}, nil
 	}
 	if !hasGlobMeta(passedFile) {
-		return []string{passedFile}, nil
+		// A plain path that is not there is a typo, not a pattern that matched nothing.
+		return nil, fmt.Errorf("%s: no such file or directory", passedFile)
```
klaernie: Could you please wrap the error returned by `os.Stat` instead of creating a new one? If the `stat` call fails for another reason - e.g. a missing permission to access the parent directory - we would misrepresent the error towards the user and send them on a wild goose chase.

pkg/files/files_test.go
```diff
@@ -229,21 +235,19 @@ func TestGetFilesGlobPattern(t *testing.T) {
...
+func TestGetFilesMissingPathWithoutMetaErrors(t *testing.T) {
 	cfg := config.NewConfig(nil)
 	cfg.PassedFiles = []string{"./does/not/exist.txt"}
 
-	files, err := GetFiles(*cfg)
-	if err != nil {
-		t.Fatalf("GetFiles(missing): unexpected error %s", err.Error())
+	// A path the user named explicitly, with no glob metacharacters, is a typo
+	// rather than a pattern that matched nothing, so it has to be an error.
+	// Previously it was passed on and silently dropped, and the run exited 0.
```
klaernie: I do not think last line helps future readers understand the context. I would only document it as it is, not reference previous implementation details

klaernie (changes requested):

VXNCXNX:
Both done, thanks.

The stat error is now wrapped, so a permission denied on the parent directory reports itself rather than claiming the file is missing:

```
./does/not/exist.txt   -> resolving ./does/not/exist.txt: stat ./does/not/exist.txt: no such file or directory
./locked/inner/f.txt   -> resolving ./locked/inner/f.txt: stat ./locked/inner/f.txt: permission denied
```

Both exit 1. `%w` keeps the sentinel, so `errors.Is(err, fs.ErrNotExist)` still works for callers that want to tell the two apart.

Dropped the line about the previous implementation from the test comment.

klaernie (approved):

theoludwig (approved):

VXNCXNX:
Fair, and you are right. My answer: error out.

That is what the branch does now, since the commit after your question. The `exit 0` you quoted was from the original before/after in the description, which I should have updated:

```
$ ec ./does/not/exist.txt
resolving ./does/not/exist.txt: stat ./does/not/exist.txt: no such file or directory
$ echo $?
1
```

A path the user typed explicitly is a typo, not an empty glob, so exiting 0 hides it. Globs that match nothing still pass through silently, since that is normal.

klaernie:
> Fair, and you are right. My answer: error out.
> 
> That is what the branch does now, since the commit after your question. The `exit 0` you quoted was from the original before/after in the description, which I should have updated:
> 
> ```
> $ ec ./does/not/exist.txt
> resolving ./does/not/exist.txt: stat ./does/not/exist.txt: no such file or directory
> $ echo $?
> 1
> ```
> 
> A path the user typed explicitly is a typo, not an empty glob, so exiting 0 hides it. Globs that match nothing still pass through silently, since that is normal.

What did you want to tell us, or did your AI agent think it still needed to reply?

VXNCXNX:
That one was noise, sorry. You had both already approved and I was answering a question that was no longer open.

PR #600, opened 2026-08-15
fix: correct the -disable-end-of-line flag description

VXNCXNX:
## What's broken

`-disable-end-of-line` documents itself as disabling a different check.

```
$ ec --help
  -disable-end-of-line
        disables the trailing whitespace check
```

That is the description of `-disable-trim-trailing-whitespace`, two entries below. The flag itself works correctly, only its description is wrong, so anyone reading `--help` to turn off the end-of-line check has no way to tell it is the right flag.

## The fix

One string in `cmd/editorconfig-checker/main.go`. The same wrong text is copied into `README.md` (the pasted `--help` output) and `docs/editorconfig-checker.1`, so both are corrected to match.

The man page is hand-maintained rather than generated, so it needed the same edit.

## Verification

A test that looks the flag up in `flag.CommandLine` and asserts its description mentions the end-of-line check and not trailing whitespace. With the old string restored it fails on both assertions. `go test ./...` passes.

### DIFF
--- a/cmd/editorconfig-checker/main_test.go
+++ b/cmd/editorconfig-checker/main_test.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"bytes"
+	"flag"
 	"os"
 	"runtime"
 	"strings"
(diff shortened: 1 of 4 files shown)


theoludwig (approved):
Thanks! :smile:

cmd/editorconfig-checker/main_test.go
klaernie, line 1: I don't think adding a test for this is really warranted.
This error was a human oversight and is unlikely to be repeated exactly. And should one assume that will happen more often and widespread, one should add tests for every option - which would be overkill.

I'd rather spend the same cpu time in code that keeps the manpage and readme portion in sync with the code.

VXNCXNX: Fair, dropped it. The PR is now the three description lines only, code, README and manpage.

Agreed on the sync check being the better use of the cycles: a test per option would not have caught this one either, since the string was wrong but well-formed. Something diffing the flag usages against the README and manpage blocks would catch the whole class, including the drift that is already there when someone edits one and not the others.

VXNCXNX:
Heads-up on the red `lint-commit`: it is my second commit message, `Drop the flag-description test per review`, which has no conventional-commits prefix. It is not the code.

The PR content is what you asked for, three description lines only:

```
 README.md                        | 2 +-
 cmd/editorconfig-checker/main.go | 2 +-
 docs/editorconfig-checker.1      | 2 +-
```

Since this repo squash-merges, that commit message is discarded at merge and the squash subject is yours to write, so it does not reach `main` either way. I would rather not force-push over the branch you have already reviewed. Say the word if you want it rewritten anyway and I will do it.

klaernie:
I think in this case even the strongest advocates against would have a hard time arguing that it becomes harder to review. Feel free to rewrite your commit message. And for the future: we have a pre-commit config that we probably don't advertise well enough - it wouldn't have let you commit this message and asked you to rewrite.

Process wise I think this PR is good to go, we'll just have our formal 24h cooldown to see if somebody has any complaints - which I doubt.

VXNCXNX:
Fixed the commit message so lint-commit passes. The run needs your approval to start.

klaernie (approved):
Thanks!

mstruebing (approved):

PR #601, opened 2026-08-15
fix: consolidate error blocks that are interleaved with other errors

### LINKED ISSUE #373: Consolidated errors when they occur in an interleaved pattern
Essentially rework the errors.go function `ConsolidateErrors` function to make the `TestConsolidatingInterleavedErrors` test pass

Eljees:
Fixes #373.

`ConsolidateErrors` walked the error list in input order and merged an error into the previous one only when the two were **adjacent in that list**. A block is therefore only recognised while no other kind of error is reported on the same lines. That is exactly what happens on a file that is both indented with tabs and carries trailing whitespace:

    a·
    →b·
    →c·
    →d·
    e

with `indent_style = space` and `trim_trailing_whitespace = true`:

    # before
    interleaved.txt:
        1-2: Trailing whitespace
        2: Wrong indent style found (tabs instead of spaces)
        3: Trailing whitespace
        3: Wrong indent style found (tabs instead of spaces)
        4: Trailing whitespace
        4: Wrong indent style found (tabs instead of spaces)

    6 errors found

    # after
    interleaved.txt:
        1-4: Trailing whitespace
        2-4: Wrong indent style found (tabs instead of spaces)

    2 errors found

Errors are now grouped by message first, then runs of consecutive lines are collected within each group. Because a map has no order, the result is sorted by line number and then by message, so the output is stable — the test file already warned about this:

> If the implementation does not sort, this test will randomly fail when the implementation uses a map

### One row of the expected table cannot be satisfied

`TestConsolidatingInterleavedErrors` expects

    {LineNumber: 4, AdditionalIdenticalErrorCount: 1, Message: errors.New("message kind 4")}

but `message kind 4` appears exactly once in the input, on line 4. `AdditionalIdenticalErrorCount` counts the *further* consecutive lines carrying the same message — `TestConsolidateErrorCounts` fixes that meaning, and the formatters print the range as `LineNumber`–`LineNumber+count`. A count of 1 would therefore print `4-5: message kind 4`, and line 5 carries `message kind 1`, not `message kind 4`.

I changed that one value to `0` and left the other five rows exactly as written. Everything else in the table is reproduced as-is:

| | expected | produced |
| --- | --- | --- |
| `-1` file-level error | 0 | 0 |
| line 1, `message kind 2` | 2 | 2 |
| line 2, `message kind 1` | 1 | 1 |
| line 3, `message kind 3` | 0 | 0 |
| line 4, `message kind 4` | 1 | **0** |
| line 5, `message kind 1` | 0 | 0 |

If the intent was for `message kind 4` to appear on lines 4 and 5, adding that second input row instead of changing the expectation works equally well and I am happy to do it that way — say the word.

### Measured

- `go test ./...`: **10 packages ok, 0 FAIL**, before and after.
- `TestConsolidateErrors` and `TestConsolidateErrorCounts` — the two tests that pin the existing behaviour — pass unchanged.
- `TestConsolidatingInterleavedErrors` runs for the first time; the `t.Skip` is removed.
- Regression check on a corpus of 11 937 files (this repository, three unrelated Python projects and the Go standard library source tree): **524 errors before, 524 after** — grouping changes how errors are presented, not which ones are found.
- `gofmt -l` and `go vet` clean. The commit follows the conventional-commit format `commitlint` enforces here.

AI-assisted (LLM used for drafting); the measurements above are mine.

### DIFF
--- a/pkg/error/error.go
+++ b/pkg/error/error.go
@@ -56,6 +56,15 @@ func byNumber(error1 ValidationError, error2 ValidationError) int {
 	return cmp.Compare(error1.LineNumber, error2.LineNumber)
 }
 
+// byNumberAndErrorMessage orders validation errors by the line they were
+// reported on, and errors sharing a line by their message.
+func byNumberAndErrorMessage(error1 ValidationError, error2 ValidationError) int {
+	if order := cmp.Compare(error1.LineNumber, error2.LineNumber); order != 0 {
+		return order
+	}
+	return cmp.Compare(error1.Message.Error(), error2.Message.Error())
+}
+
 func ConsolidateErrors(errors []ValidationError, config config.Config) []ValidationError {
 	var lineLessErrors []ValidationError
 	var errorsWithLines []ValidationError
@@ -106,12 +115,7 @@ func ConsolidateErrors(errors []ValidationError, config config.Config) []Validat
 
 	// map iteration order is random, so put the result back into a stable,
 	// human-friendly order: by line, then by message
-	slices.SortStableFunc(consolidatedErrors, func(a, b ValidationError) int {
-		if c := cmp.Compare(a.LineNumber, b.LineNumber); c != 0 {
-			return c
-		}
-		return cmp.Compare(a.Message.Error(), b.Message.Error())
-	})
+	slices.SortStableFunc(consolidatedErrors, byNumberAndErrorMessage)
 
 	return append(lineLessErrors, consolidatedErrors...)
 }
--- a/pkg/error/error.go
+++ b/pkg/error/error.go
@@ -51,6 +51,11 @@ func GetErrorCount(errors []ValidationErrors) int {
 	return errorCount
 }
 
+// byNumber orders validation errors by the line they were reported on.
+func byNumber(error1 ValidationError, error2 ValidationError) int {
+	return cmp.Compare(error1.LineNumber, error2.LineNumber)
+}
+
 func ConsolidateErrors(errors []ValidationError, config config.Config) []ValidationError {
 	var lineLessErrors []ValidationError
 	var errorsWithLines []ValidationError
@@ -81,9 +86,7 @@ func ConsolidateErrors(errors []ValidationError, config config.Config) []Validat
 	for message, groupErrors := range grouped {
 		config.Logger.Debug("consolidating %d errors with message %q", len(groupErrors), message)
 
-		slices.SortStableFunc(groupErrors, func(a, b ValidationError) int {
-			return cmp.Compare(a.LineNumber, b.LineNumber)
-		})
+		slices.SortStableFunc(groupErrors, byNumber)
 
 		for i := 0; i < len(groupErrors); i++ {
 			thisError := groupErrors[i]
--- a/pkg/error/error.go
+++ b/pkg/error/error.go
@@ -99,7 +99,6 @@ func ConsolidateErrors(errors []ValidationError, config config.Config) []Validat
 
 		for i := 0; i < len(groupErrors); i++ {
 			thisError := groupErrors[i]
-			thisError.AdditionalIdenticalErrorCount = 0
 
 			// walk forward while the next error sits on the line right after
 			// the range collected so far
--- a/pkg/error/error.go
+++ b/pkg/error/error.go
@@ -79,11 +79,11 @@ func ConsolidateErrors(errors []ValidationError, config config.Config) []Validat
 	var consolidatedErrors []ValidationError
 
 	for message, groupErrors := range grouped {
+		config.Logger.Debug("consolidating %d errors with message %q", len(groupErrors), message)
+
 		slices.SortStableFunc(groupErrors, func(a, b ValidationError) int {
 			return cmp.Compare(a.LineNumber, b.LineNumber)
 		})
-
-		config.Logger.Debug("consolidating %d errors with message %q", len(groupErrors), message)
 
 		for i := 0; i < len(groupErrors); i++ {
 			thisError := groupErrors[i]
--- a/pkg/error/error.go
+++ b/pkg/error/error.go
@@ -66,26 +66,12 @@ func byNumberAndErrorMessage(error1 ValidationError, error2 ValidationError) int
 }
 
 func ConsolidateErrors(errors []ValidationError, config config.Config) []ValidationError {
-	var lineLessErrors []ValidationError
-	var errorsWithLines []ValidationError
-
-	// filter the errors, so we do not need to care about LineNumber == -1 in the loop below
-	for _, singleError := range errors {
-		if singleError.LineNumber == -1 {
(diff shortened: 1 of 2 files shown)


pkg/error/error.go
```diff
@@ -64,26 +66,50 @@ func ConsolidateErrors(errors []ValidationError, config config.Config) []Validat
...
-				thisError.AdditionalIdenticalErrorCount++ // keep track of how many consecutive lines we've seen
-				i++                                       // make sure the outer loop jumps over the consecutive errors we just found
-			} else {
-				break // if they are different errors we can stop comparing messages
+	for message, groupErrors := range grouped {
+		slices.SortStableFunc(groupErrors, func(a, b ValidationError) int {
+			return cmp.Compare(a.LineNumber, b.LineNumber)
+		})
+
+		config.Logger.Debug("consolidating %d errors with message %q", len(groupErrors), message)
```
klaernie, line 82: I would move this before the sorting of the error group - doesn't make sense to me to log this after already having started the work

pkg/error/error.go
```diff
@@ -64,26 +66,50 @@ func ConsolidateErrors(errors []ValidationError, config config.Config) []Validat
...
-			config.Logger.Debug("comparing against error %d(%s)", j, nextError.Message)
-			if nextError.Message.Error() == thisError.Message.Error() && nextError.LineNumber == thisError.LineNumber+thisError.AdditionalIdenticalErrorCount+1 {
-				thisError.AdditionalIdenticalErrorCount++ // keep track of how many consecutive lines we've seen
-				i++                                       // make sure the outer loop jumps over the consecutive errors we just found
-			} else {
-				break // if they are different errors we can stop comparing messages
+	for message, groupErrors := range grouped {
+		slices.SortStableFunc(groupErrors, func(a, b ValidationError) int {
+			return cmp.Compare(a.LineNumber, b.LineNumber)
+		})
```
klaernie: could you extract the comparator function into a function named `byNumber()`? This is more because of the second one later down.

pkg/error/error.go
```diff
@@ -64,26 +66,50 @@ func ConsolidateErrors(errors []ValidationError, config config.Config) []Validat
...
 	}
 
+	// map iteration order is random, so put the result back into a stable,
+	// human-friendly order: by line, then by message
+	slices.SortStableFunc(consolidatedErrors, func(a, b ValidationError) int {
+		if c := cmp.Compare(a.LineNumber, b.LineNumber); c != 0 {
+			return c
+		}
+		return cmp.Compare(a.Message.Error(), b.Message.Error())
+	})
```
klaernie: Could you please extract this comparator function into a function named `byNumberAndErrorMessage()`? This way we can add a unit test for it - which should then also cover the second order comparison on `.Error()`.

pkg/error/error.go
```diff
@@ -64,26 +66,50 @@ func ConsolidateErrors(errors []ValidationError, config config.Config) []Validat
 
 	config.Logger.Debug("sorted errors: %d with line number -1, %d with a line number", len(lineLessErrors), len(errorsWithLines))
```
klaernie: I think with sorting by error first, and sorting the final result by line number first, it isn't necessary to treat the errors without a line number separately, since the comparison became much easier. This probably wouldn't have been to hard with the original implementation, but I wasn't far enough in my golang journey to see the wood for the trees when I built this originally.

pkg/error/error.go
```diff
@@ -64,26 +66,50 @@ func ConsolidateErrors(errors []ValidationError, config config.Config) []Validat
...
+	for message, groupErrors := range grouped {
+		slices.SortStableFunc(groupErrors, func(a, b ValidationError) int {
+			return cmp.Compare(a.LineNumber, b.LineNumber)
+		})
+
+		config.Logger.Debug("consolidating %d errors with message %q", len(groupErrors), message)
+
+		for i := 0; i < len(groupErrors); i++ {
+			thisError := groupErrors[i]
+			thisError.AdditionalIdenticalErrorCount = 0
```
klaernie: `.AdditionalIdenticalErrorCount` is always 0 by default, and if a future validator would ever produce already-consolidated errors (where the validator scans the entire file on one go) it would break this validators output.

klaernie (changes requested):
Great work, thank you very much!

I have a few minor details I'd like to changed, but in terms of functionality this looks very good!

Eljees:
Thanks — all five are fair, I will take them all.

1. Debug line moves above the sort.
2. / 3. Both comparators become named functions, `byNumber()` and `byNumberAndErrorMessage()`, with a unit test on the second one that covers the second-order comparison on `.Error()`.
5. The `AdditionalIdenticalErrorCount = 0` reset goes away. You are right that the reset is destructive: a validator that scans a whole file at once and hands back an already-consolidated error would have its count silently zeroed.

On 4: agreed, the split can go. Errors with `LineNumber == -1` can run through the same grouping, because the block-continuation test is `next.LineNumber == this.LineNumber + count + 1`, which for two line-less errors is `-1 == 0` and therefore never merges them. They still sort to the top, since `-1 < 1`.

One behavioural detail so it is not a surprise later: today line-less errors keep their input order, and after the change several *different* file-level messages on one file would be ordered by message text. No current test covers that case — `TestConsolidateErrors`, `TestConsolidatingInterleavedErrors` and `TestFormatErrors` each use a single line-less error per file — so nothing goes red either way. Say if you would rather keep input order for those; otherwise I will take the simpler comparison.

Will push the four changes and drop the split in one commit.

klaernie:
> Will push the four changes and drop the split in one commit.

I also wouldn't mind seeing four individual commits - it makes reviewing much easier if the changes are not a big amalgamation but concise with a commit message describing it.

> today line-less errors keep their input order, and after the change several _different_ file-level messages on one file would be ordered by message text.

Since the order currently is only determined by the order in which they are done in the code, I don't think this is a problem for anyone. I think the complexity of trying to preserve the order is not needed.

Eljees:
Pushed as five commits, one per point.

| commit | |
| --- | --- |
| `refactor: log the group before sorting it` | the debug line now runs before the sort |
| `refactor: extract the line-number comparator as byNumber` | |
| `refactor: extract byNumberAndErrorMessage and cover it with a test` | six cases, including both directions of the message tie-break and the equal case |
| `fix: keep a count a validator already reported` | the `AdditionalIdenticalErrorCount = 0` reset is gone |
| `refactor: run file-level errors through the same grouping` | the `LineNumber == -1` split is gone |

On the last one — I took your reading and did not try to preserve input order for file-level errors. Several different file-level messages on one file now come out ordered by message text.

The branch is rebased onto current main, so the range is clean.

`go test ./...`: 10 packages ok, 0 FAIL — and the suite is green after each of the five commits individually, not only at the end. `gofmt -l` and `go vet` are clean.

klaernie (approved):

theoludwig (approved):

PR #605, opened 2026-08-17
chore: disable commitlint footer-max-line-length rule

theoludwig:
Same reasoning as to why `body-max-line-length` is disabled, I think we should also disable the `footer-max-line-length` rule for commit message.

The goal of git commit message is to describe and have context of why the changes were made, restricting it makes it harder to describe.

Opened this PR because my PRs #603 and #604 started failing CI because of commitlint.


klaernie (approved):

mstruebing (approved):
Totally fine with that

PR #606, opened 2026-08-17
feat: auto-detect `github-actions` output format in CI

### LINKED ISSUE #596: Auto-detect CI
Consider using the existing `github.com/gkampitakis/ciinfo` dependency to auto-detect CI environments. 

For example, when no format is explicitly provided, set the output format to `github-actions` if `ciinfo.GITHUB_ACTIONS` is true. 

Explicit CLI and configuration values should take precedence, preserving the current behavior outside CI while avoiding the need to manually configure `--format github-actions` in GitHub Actions workflows.

One reason is that often linting tasks are defined in a single place (e.g. via mise) running in both CI and on developer machines, and in those cases setting via a flag or config file doesn't work. But auto-detect does! 😄

theoludwig:
Fixes #596


klaernie (approved):

cmd/editorconfig-checker/main.go
```diff
@@ -10,6 +10,8 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/gkampitakis/ciinfo"
```
mstruebing, line 13: Do we need this dependency?
Couldn't we just have a look if the `GITHUB_ACTION` environment variable is set or alternatively `CI`?  
https://docs.github.com/en/actions/reference/workflows-and-actions/variables

klaernie: We already have this dependency, but it was an indirect one through go-snaps.

I would also imagine, that in the not so far future somebody else tells us about a relation of a CI platform and it's preferred output format (I think GitLab and codeclimate might be one, but I haven't yet found the proof from where I think I've got this information), at which point we would need to implement another one ourselves. I'd prefer using a module which is better tested that we could do on our own.

mstruebing: Alright you've convinced me.

mstruebing (approved):

sandstrom:
Great! 💯

PR #609, opened 2026-08-18
chore: match Go files by glob in release-please extra-files

theoludwig:
## Problem

The 4.0.0 release PR (#595) fails CI:

```
pkg/files/files_test.go:15:2: no required module provides package
github.com/editorconfig-checker/editorconfig-checker/v3/pkg/config
```

`pkg/files/files_test.go` was missing from `extra-files` in `release-please-config.json`.

(still using "v3" path instead of v4)

## Fix

Rather than just adding the missing line, I thought it would be a good idea to replace by a single glob to avoid further mistakes.

## Note for after this merges

This fixes future releases but does not repair #595, since that branch already has the bad v3 line committed.

The `release-please--branches--main` branch should be deleted after this lands so the release PR is regenerated from scratch.

### DIFF
--- a/release-please-config.json
+++ b/release-please-config.json
@@ -16,16 +16,11 @@
                 "README.md",
                 "go.mod",
                 "default.nix",
-                "cmd/editorconfig-checker/main.go",
-                "pkg/config/config.go",
-                "pkg/config/config_test.go",
-                "pkg/error/error.go",
-                "pkg/error/error_test.go",
-                "pkg/files/files.go",
-                "pkg/validation/validation.go",
-                "pkg/validation/validation_test.go",
-                "pkg/validation/validators/validators.go",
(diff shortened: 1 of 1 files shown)


klaernie (approved):
Turns out the depth of release-please's feature is much more than they ever documented in prose.

I'll merge this immediately, but I don't think we need to recreate the v4 release PR - release-please regenerates it again and again, always rebasing it one main. So I would expect the PR to be updated and the remaining file to be caught.

theoludwig:
> Turns out the depth of release-please's feature is much more than they ever documented in prose.
> 
> I'll merge this immediately, but I don't think we need to recreate the v4 release PR - release-please regenerates it again and again, always rebasing it one main. So I would expect the PR to be updated and the remaining file to be caught.

Yep, you we were right, no need to create again the branch. :smile:

PR #612, opened 2026-08-19
feat: add safe EditorConfig autofixes

### LINKED ISSUE #14: Autofix flag
introduce an autofix flag which should fix all issues which can be fixed.

zeitlinger:
## Summary

Add an opt-in `--fix` mode that applies a deliberately small set of safe EditorConfig fixes before validation.

This starts a fresh implementation based on the supportive discussion in [issue #14](https://github.com/editorconfig-checker/editorconfig-checker/issues/14), rather than reviving the older [pull request #61](https://github.com/editorconfig-checker/editorconfig-checker/pull/61). The fresh implementation narrows the scope and documents the safety boundary explicitly.

## Supported fixes

`--fix` currently applies only:

- `trim_trailing_whitespace = true` (while honoring line-level disable directives)
- `end_of_line = lf`, `cr`, or `crlf`
- `insert_final_newline = true` or `false`

Indentation, `indent_size`, `max_line_length`, and `charset` remain check-only. Unknown rule values are reported by normal validation and are never guessed during fixing.

## Safety and file semantics

- Files are checked before modification and only regular, non-empty, non-symlink files are eligible.
- Binary-like content, invalid UTF-8, and unsupported encodings are skipped rather than rewritten.
- UTF-8 BOMs are preserved.
- File permissions, including Unix special permission bits, are retained.
- File-level disable directives and applicable line-level disable directives are honored.
- Content is written through a same-directory temporary file and replacement. This is crash-resistant and atomic where the operating system/filesystem provides those guarantees; hard links to the old inode are not updated.
- A second `--fix` run is idempotent.

The original file is revalidated immediately before replacement. If the path, inode, size, or modification time changed while fixing, replacement is refused rather than overwriting concurrent edits. Fix errors return the existing error exit status; after successful fixes, the normal validation pipeline still runs and determines the final exit status.

## Validation

- `make test`
- `go test -race ./...`
- `make build`
- `GOOS=windows GOARCH=amd64 go build -o /tmp/editorconfig-checker-windows-amd64.exe ./cmd/editorconfig-checker`
- `git diff --check`

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -128,7 +128,7 @@ type Config struct {
 	ShowVersion bool
 	Help        bool
 	DryRun      bool
-	Fix         bool `json:"-"`
+	Fix         bool
 	Path        string
 
 	// CONFIG FILE
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -128,6 +128,7 @@ type Config struct {
 	ShowVersion bool
 	Help        bool
 	DryRun      bool
+	Fix         bool `json:"-"`
 	Path        string
 
 	// CONFIG FILE
@@ -217,6 +218,10 @@ func (c *Config) Parse() error {
 func (c *Config) Merge(config Config) {
 	if config.DryRun {
 		c.DryRun = config.DryRun
+	}
+
+	if config.Fix {
+		c.Fix = config.Fix
 	}
 
 	if config.ShowVersion {
--- a/pkg/fix/fix.go
+++ b/pkg/fix/fix.go
@@ -21,77 +21,84 @@ import (
 // not modified: rewriting those files without an encoding-aware encoder could
 // corrupt their contents.
 func FixFile(filePath string, cfg config.Config, def *editorconfig.Definition) (bool, error) {
+	info, content, ok, err := readFixableFile(filePath, cfg)
+	if err != nil || !ok {
+		return false, err
+	}
+	fixed, changed := fixContent(content, cfg, def)
+	if !changed {
+		return false, nil
+	}
+	if err := atomicWrite(filePath, fixed, fileMode(info), info); err != nil {
+		return false, err
+	}
+	return true, nil
+}
+
+func readFixableFile(filePath string, cfg config.Config) (os.FileInfo, []byte, bool, error) {
 	info, err := os.Lstat(filePath)
 	if err != nil {
-		return false, err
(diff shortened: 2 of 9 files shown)


pkg/config/config.go
```diff
@@ -128,6 +128,7 @@ type Config struct {
 	ShowVersion bool
 	Help        bool
 	DryRun      bool
+	Fix         bool `json:"-"`
```
mstruebing: Should we remove that `json` thingy?

pkg/fix/fix.go
```diff
@@ -0,0 +1,302 @@
...
+	"github.com/editorconfig-checker/editorconfig-checker/v3/pkg/config"
+	"github.com/editorconfig-checker/editorconfig-checker/v3/pkg/encoding"
+)
+
+// FixFile applies the safe, supported fixes for filePath. It returns true when
+// the file was changed. Rules which are not explicitly supported remain check-only.
+// Files which are not valid UTF-8 (including binary-like files) are deliberately
+// not modified: rewriting those files without an encoding-aware encoder could
+// corrupt their contents.
+func FixFile(filePath string, cfg config.Config, def *editorconfig.Definition) (bool, error) {
```
mstruebing, line 23: Should this function be split up into multiple? Maybe per type of fix (eol, trailing whitespace, ...)

pkg/config/config.go
zeitlinger: Removed the unnecessary `json:"-"` tag from `Config.Fix`, keeping it consistent with the other CLI fields. Added merge coverage and updated the `Config.String()` snapshot so the field’s JSON behavior is covered.

pkg/fix/fix.go
zeitlinger: Refactored `FixFile` into focused safety/encoding, policy parsing, line-fixing, final-newline, and orchestration helpers while preserving behavior. Added coverage for EOL/final-newline policies, directives, unsafe files, CLI success/error/verbose paths, and atomic lost-update protection.

klaernie (approved):
I thoroughly looked at the proposed changes and it seems all quite reasonable to me. Since this is a new opt-in feature I think we could merge it as is and declare it a beta feature, asking users to provide feedback - I would assume there are a few edge cases out in the wild that I simply could not think of.

theoludwig (approved):

PR #614, opened 2026-08-25
build(deps): update Go to v1.27.0

theoludwig:
Fixes #613

PR #615, opened 2026-08-28
fix: allow builds with Go 1.26

twpayne:
#614 bumped the the `go` version in `go.mod` from `go 1.25.0` to `go 1.27.0`. This specifies that the minimum version of Go required to build `editorconfig-checker` is now 1.27 ([ref](https://go.dev/ref/mod#go-mod-file-go)). However, this is not correct: `editorconfig-checker` builds just fine with Go 1.26.

The downside of the `go 1.27.0` directive is the everything that depends on `editorconfig-checker` now must be built with minimum Go 1.27 as well, effectively breaking `editorconfig-checker` for any project that uses it via a `tool` directive and wanting to support an older version of Go.

This PR changes the minimum version of Go to 1.26.0, which is a reasonable default since Go 1.26.x is the oldest version of Go supported by the Go team.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,7 +27,7 @@ jobs:
           fetch-depth: 0
       - uses: actions/setup-go@v7
         with:
-          go-version-file: "go.mod"
+          go-version: stable
       - run: make build
       - run: make test
       - run: make run
@@ -43,7 +43,7 @@ jobs:
       - uses: actions/checkout@v7
       - uses: actions/setup-go@v7
         with:
-          go-version-file: "go.mod"
+          go-version: stable
       - run: make test
       - run: make run
       - run: make bench
@@ -58,7 +58,7 @@ jobs:
       - uses: actions/checkout@v7
       - uses: actions/setup-go@v7
         with:
-          go-version-file: "go.mod"
+          go-version: stable
       - run: make test
       - run: make run
       - run: make bench
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -25,7 +25,7 @@ jobs:
       - run: git fetch --force --tags
       - uses: actions/setup-go@v7
         with:
-          go-version-file: "go.mod"
+          go-version: stable
           cache: true
 
       - name: Set up QEMU
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/editorconfig-checker/editorconfig-checker/v3 // x-release-please-major
 
-go 1.27.0
+go 1.26.0
 
 require (
 	github.com/editorconfig/editorconfig-core-go/v2 v2.6.4


klaernie (approved):
That's definitely a case I haven't thought of, but I'm also not a big fan of [the tool directive](https://go.dev/doc/go1.24#tools). I recently removed a pre-go1.24 DIY implementation at work, since the CI needed to always build the tools from scratch with each run, and installing goreleaser and mage using their binaries was a way more efficient on our compute and network resources. I however don't know if that problem also applies to `go tool`.

Generally we probably should adopt the same supported go version strategy as golang itself - latest and latest-1. We'd just need to figure out how to tell renovate this - in my cursory search I haven't found something particularly helpful yet.

theoludwig (dismissed):
Go is backwards compatible, so you should be able to upgrade to 1.27.0 without problems, and we're preparing v4.0.0 in #595 so IMO, it's the right thing to do, since we will never update it again after for a very long time (multiple years between each `editorconfig-checker` major version), if we need to consider that kind of change as a breaking change.

(but tbh, I didn't know that field was the "minimum go version" sorry for breaking your workflow, I thought it was only meaning which go version we use to build it ourselves, thanks for the explanations)

twpayne:
Yes, Go is backwards compatible, but that doesn't mean that everybody can upgrade.

[My project](https://github.com/twpayne/chezmoi) runs on every system supported by Go. This includes platforms like FreeBSD and OpenBSD which often take several months to upgrade their Go major versions. So, if I want to support them I have to ensure that my code builds with the old stable version of Go, i.e. 1.26.

As a short-term workaround, I've blacklisted the latest version of `editorconfig-checker` that requires Go 1.27 (3.11.2), but I'd prefer not to have to do this.

twpayne:
Note that listing a more recent version of Go than you actually need in `go.mod` has no upside and only downsides. You control which version of Go is used to build your release binaries in your CI config, not in your `go.mod` file.

klaernie:
> since we will never update it again after for a very long time (multiple years between each `editorconfig-checker` major version), if we need to consider that kind of change as a breaking change

I don't think we need to consider it a breaking change.
IMHO there are two hearts beating in us:
- the first wants to ensure that we build our binaries using the latest version of golang available, so the most amount of bugs are fixed, the number of CVEs is lowest and we get the performance benefits that a newer version might bring.
- and the second one wants to ensure that users like @twpayne can use any mechanism they like to fetch our latest version, without having them forced to upgrade to the latest golang until they decide themselves. However we need to have a certain boundary on how old a version of golang we can and want to support. At the moment the newest feature I'm aware of is `testing.T.Chdir()` introduced with go1.24.0. And looking at our dependencies github.com/editorconfig/editorconfig-core-go/v2 is the first module I found that require a newer version - namely 1.25. And for making up a sensible policy I'd suggest: "we support the latest stable version of golang and the version before".

Putting this together we'd use `go-version: stable` in our CI instead of the current `go-version-file: go.mod`, and merge this PR. This way we have the latest and greatest for the binary and can support users like @twpayne.

What do y'all think?

twpayne:
> Putting this together we'd use `go-version: stable` in our CI instead of the current `go-version-file: go.mod`, and merge this PR. This way we have the latest and greatest for the binary and can support users like @twpayne.

For me, this is the perfect outcome: `editorconfig-checker` gets built with the latest version of Go, while still supporting as many projects as possible.

theoludwig:
Thanks both for the explanations.

> Putting this together we'd use `go-version: stable` in our CI instead of the current `go-version-file: go.mod`, and merge this PR. This way we have the latest and greatest for the binary and can support users like @twpayne.

Sounds good. :+1:

Could you please update the PR so that we use Go v1.26 as the minimum requirement but then v1.27 for building our binaries when releasing?

and if we really want to do it well, then we would need to update our CI to tests both v1.26 and v1.27, since we support both, but that's probably not needed or worth the complexity.

___

Slightly off-topic hot take: I don't like the philosophy of software where people take forever to update and let things rot.
@klaernie already listed the advantages "bugs are fixed, the number of CVEs is lowest and we get the performance benefits" and more.

But as a developer, I'd be quite mad if people never used what I built because they're afraid of breakage. Of course updating means some things may break or behave differently, but then report those issues to the maintainers.

Some devs work so hard on improving their software, for people to not use it. So many companies still run old software... and it's even more depressing when it's software like Go that treats backward compatibility as a priority. It's one of the reasons I use a rolling-release distro (Arch Linux).

I'm not going to dwell on it any further. :laughing:

twpayne:
Done. I've changed the `go-version` in the Github Action workflows to use `stable` for all steps, including CI and release. This will mean that `editorconfig-checker` is always tested and built with the latest stable Go version.

theoludwig (approved):

klaernie (approved):

klaernie:
> and if we really want to do it well, then we would need to update our CI to tests both v1.26 and v1.27, since we support both, but that's probably not needed or worth the complexity.

It wouldn't be too hard - after all in #616 I switched the CI to a matrix anyway, so running CI against `stable` and `oldstable` would be trivial. 

I'll take the liberty and merge this PR now, since there is practically no risk in this PR.

---

> Slightly off-topic hot take: I don't like the philosophy of software where people take forever to update and let things rot.

I'm with you there, but there are a few legitimate cases in between. In my work we're also under pressure to always use software with less vulnerabilities, and even going so far to preferably run software compiled in house from trusted sources. But on the other hand some of our units require snapshotted linux repos to ensure a uniform base, so it doesn't matter on which day of the week an application server was patched - given the hardware is willing - they will all work and fail identically. Of cause this only means weekly and daily snapshots, not 6+ month old software (although one could argue that this applies to SLES in total).

But even in a case with *BSD or Debian I see the reason why it takes so long: upgrading something as far down the stack as a compiler means that the resources for a rebuild of the entire stack on top are enormous, and the resources to verify that all build results are working correctly and fix the bugs are hard to come by. And in these projects the upgrade looks simple at first, but there will inevitable be a shortcoming of humans testing the software to ensure it works as planned. And one surely can replace a lot of the testing with hopefully well written tests, but nobody is building integration tests for an entire distribution. Else we'd probably see much more frequent releases from Debian.

And even taking it to my work: I'd be happy to run a rolling SLES release at work, but that's because I'm very cognisant of my particular system boundary - I know how my Nagios instances will interact with the OS and kernel, and what the typical failures are. And it turns out that humans are the most problematic (particularly me writing code). But other units are much more cautious and would not accept this risk - however they run the systems running our financial statements, I just monitor them.

IMHO, putting all together we're not particularly encouraging our users now to use an older golang, but enable them if they have to. And most of our users are using a binary built by us - looking at the asset downloads there are 20k downloads in just 5 days (interestingly the compat archive), I know that super-linter has a bunch of users, and they copy the binary from our docker image.

PR #616, opened 2026-08-28
ci: run exactly the same pipeline for all OSes

klaernie:
My primary motivation for this is to run the exactly same test suite with coverage on all OSes, so things like different path separator codepaths get exercised and captured by codecov.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,7 +31,7 @@ jobs:
           - windows-latest
     runs-on: ${{ matrix.os }}
     steps:
-      - name: setup git to always use linefeed as end-of-line during checkout (windows sadly has a diofferent default)
+      - name: setup git to always use linefeed as end-of-line during checkout (windows sadly has a different default)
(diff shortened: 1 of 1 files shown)


mstruebing (approved):
nice, I think we need to change the required status checks for this.
I can do that but will leave it open for now to give others the chance to look at it.

theoludwig (approved):

.github/workflows/ci.yml
```diff
@@ -22,9 +22,20 @@ jobs:
...
+    name: CI
+    strategy:
+      matrix:
+        os:
+          - ubuntu-latest
+          - macos-latest
+          - windows-latest
+    runs-on: ${{ matrix.os }}
     steps:
+      - name: setup git to always use linefeed as end-of-line during checkout (windows sadly has a diofferent default)
```
theoludwig: typo: "diofferent"

mstruebing: I've fixed it.
I will merge once CI finishes and adjust the needed checks.

PR #617, opened 2026-08-30
ci!: migrate to goreleaser dockers_v2

klaernie:
Due to the way dockers_v2 works, this will mean we no longer publish single-arch tags:

- `3.11.2-amd64`
- `3.11.2-arm32v6`
- `3.11.2-arm64v8`
- `3.11.2-i386`

So the only tags being released in the future will be

- `3.11.2`
- `3.11`
- `3`
- `latest`

However, looking at the only statistics available to us (the per-image pulls in github) the single-arch images were used only 4 times, which exactly matches the number of multi-arch tags we are creating.

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -64,153 +64,28 @@ nfpms:
       - termux.deb
       - archlinux
 
-dockers:
+dockers_v2:
   - id: docker-linux-amd64
     ids:
       - editorconfig-checker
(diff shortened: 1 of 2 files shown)


theoludwig (approved):

mstruebing (approved):

mstruebing:
@klaernie you need to rebase after #616 merged.

PR #620, opened 2026-09-01
docs: document our decision to support oldstable go

klaernie:
This is a documentation and one tiny addition to hopefully steer users of `go install` to use the right toolchain to build (the toolchain directive only is effective, if the module is being build as the main module).

But I'd request @twpayne to verify that this works as I understood it.


twpayne (approved):
LGTM

theoludwig (approved):

mstruebing (approved):

Issue #580, opened 2026-07-08
Use OS native line ending for "insert_final_newline" check when "end_of_line = unset"


Ollie-Moss2:
## Overview:
The "insert_final_newline" check should default to using the OS native line ending when checking for a final new line when "end_of_line = unset".

## Current Behaviour:
Given this file and editorconfig I get an unexpected error:\
\
**myjsonfile.json:**
```json
{
   "somejson": "with a final crlf final new line",
}
\r\n
```

**Editor Config:**
```ini
# All files
[*]
indent_style = space
# Line endings are handled because we use text=auto in gitattributes 
end_of_line = unset
indent_size = 2
insert_final_newline = true
```

**Output:**
```
myjsonfile.json:
        Wrong line endings or no final newline
```

## Expected Behaviour:
No error to be produced.

## Proposed Fix:
Update `utils.go:GetEolChar()` to default to OS native line ending and not just `\n`.\
*I don't know Go so I am unsure exactly how this would be implemented*
```go
func GetEolChar(endOfLine string) string {
	switch endOfLine {
	case "lf":
		return "\n"

	case "cr":
		return "\r"
	case "crlf":
		return "\r\n"
	}

    // default to OS native line ending here...
        
	return "\n"
}
```

mstruebing:
I'm not 100% sure right now but isn't it implemented that when it's unset it doesn't fulfill the check at all?
If not, I think this should be the right behavior.

PR #622, opened 2026-09-03
fix: keep the file's existing line ending when adding a final newline

dylanpulver:
With `end_of_line` unset, `--fix` appends an **LF** final newline to a
consistently-CRLF file, leaving it mixed. The checker can't catch it: with
`end_of_line` unset it deliberately accepts any ending (#580, #594).

```
.editorconfig: [*] insert_final_newline = true
before:  "alpha\r\nbeta\r\ngamma"
--fix →  "alpha\r\nbeta\r\ngamma\n"    # mixed
now      "alpha\r\nbeta\r\ngamma\r\n"
```

`detectFinalEOL` only inspects the suffix, so it returns `""` in exactly the
state its caller asks about — a file with no final newline — and the caller
falls back to `"\n"`. Its non-empty return can never change the outcome: it is
non-empty only when the content already ends with that ending, and `HasSuffix`
then skips the append. So the appended ending was always LF. It now reads the
last terminator anywhere in the file.

The tests were blind: every `detectFinalEOL` row already ended with a newline.

Naive alternative (first ending found anywhere) **passes the whole suite**, so
the tests don't decide it. Over 21,844 enumerated inputs the two agree on all
6,745 with a consistent or absent ending, differing only on already-mixed
files. I took the last complete line's ending, the line the new one continues.

Ran `go test -race ./...`, `make test`, golangci-lint 2.13.1 and the binary on
this repo — all clean. Reverting the helper fails only the new assertions.

AI-assisted (Claude Code, claude-opus-5); reviewed and verified by me.

### DIFF
--- a/internal/fix/fix.go
+++ b/internal/fix/fix.go
@@ -311,17 +311,24 @@ func splitLines(content []byte) []linePart {
 	return result
 }
 
+// detectFinalEOL returns the line terminator that ends the last complete line
+// of content, or "" when content holds no line terminator at all. It looks for
+// the last terminator anywhere in content rather than only at the suffix,
+// because the caller needs an answer precisely when content does *not* end
+// with a terminator; a suffix-only check returns "" in exactly that case and
+// leaves the caller guessing.
 func detectFinalEOL(content []byte) string {
-	if bytes.HasSuffix(content, []byte("\r\n")) {
-		return "\r\n"
+	idx := bytes.LastIndexAny(content, "\r\n")
+	if idx < 0 {
+		return ""
 	}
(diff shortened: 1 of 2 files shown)


klaernie (approved):
I think this is a definitive improvement, but it also reuses the last-found line-ending.
So a file like `a\nb\nc\nd\r\ne` will be "fixed" to `a\nb\nc\nd\r\ne\r\n`.

It would however be quite compute-intensive to first iterate over the entire file, counting the different line endings to then check which line ending is the predominant one, statistically speaking.

So from my perspective I would call this good to merge.

theoludwig (approved):

PR #629, opened 2026-09-15
fix(encoding): detect 7-bit content with control characters as ASCII

### LINKED ISSUE #627: UTF-8 detected as ISO-8859-1
Hi, thanks for supporting the community with this tool.

I'm experiencing an issue which I believe is a false positive.

I have a file that contains ANSI escape sequences (such as terminal colors `\x1b[32m...`) and 7-bit ASCII characters.

editorconfig-checker v4.0.1 detects this file's encoding as `ISO-8859-1` instead of `utf-8`.

`editorconfig-checker` fails with:

```text
wrong character encoding ("ISO-8859-1" instead of "utf-8")
```

## Steps to Reproduce

1. Create a minimal .editorconfig:

   ```ini
    root = true

    [*]
    charset = utf-8
    ```

2. Create a file containing ANSI color escape sequences:
 
    ```
    printf '\x1b[32mHello World\x1b[0m\n' > test-ansi.txt
    ```

3. Run editorconfig-checker:
 
    ```
    editorconfig-checker test-ansi.txt
    ```

## Expected Behavior

The file should pass validation with exit code `0`.

Every byte in test-ansi.txt is strictly < 0x80 (ASCII range), which makes it valid UTF-8 per RFC 3629.

The escape character (0x1B) is a standard 7-bit ASCII/C0 control character, not an indicator of ISO-8859-1.

## Actual Behavior

editorconfig-checker reports:

```
test-ansi.txt:
    wrong character encoding ("ISO-8859-1" instead of "utf-8")

1 errors found
```

Thanks for the support!

anandghegde:
Fixes #627.

### The bug

A file that is plain 7-bit ASCII apart from a C0 control character is reported as `ISO-8859-1`, so `charset = utf-8` fails on it. The reporter's case is a text file holding ANSI color codes, whose only non-printable byte is ESC (`0x1b`):

```console
$ printf '\033[32mHello World\033[0m\n' > colored.txt
$ cat .editorconfig
root = true

[*]
charset = utf-8

$ editorconfig-checker .
colored.txt:
	wrong character encoding ("ISO-8859-1" instead of "utf-8")

1 errors found
```

### Cause

In `Detect`, the ASCII branch requires the content to hold no C0 character at all:

```go
if !hasC0 && !hasC1 && !hasHI {
    encoding = consts.Ascii
    break
}

if !hasHI {
    encoding = consts.ISO88591
    break
}
```

ESC is in `c0Chars`, so `hasC0` is true and the file falls through to the second branch. `hasC1` is redundant there, since `c1Chars` is a subset of `hiChars`, so that branch is reached only for content that is valid UTF-8, has no byte `>= 0x80`, and holds a C0 character. All of that content is 7-bit, which makes `ISO-8859-1` the wrong answer for every file that reaches it.

### Fix

Report ASCII whenever no byte is `>= 0x80`, control characters included. Since 7-bit content is a strict subset of UTF-8, `CharsetsMatch` already lets `ascii` satisfy `charset = utf-8`.

ISO-2022 family encodings are 7-bit as well, so a positive detection of one of those is kept, guarded on a decoder being available:

```go
if !hasHI {
    if !isDecodableISO2022Encoding(encoding) {
        encoding = consts.Ascii
    }
    break
}
```

That guard also fixes a second, smaller problem: the old branch overwrote chardet's correct ISO-2022-JP detection (99% confidence) with `ISO-8859-1`. The decoder condition keeps `ISO-2022-KR` and `ISO-2022-CN`, which have no decoder in this package, from being reported and then falling back to `binary`.

### Tests

New fixture `testdata/issues/issue627.txt` is the reporter's example, with an entry in `test-results.json` expecting `Ascii`. It fails on `main` (`got ISO-8859-1, want Ascii`) and passes here.

Seven existing expectations in `test-results.json` change, all of them for the better:

| file | before | after |
| --- | --- | --- |
| `testdata/uchardet/en/ascii.txt` | ISO-8859-1 | Ascii |
| `testdata/text/rashomon-iso-2022-jp.txt` | ISO-8859-1 | ISO-2022-JP |
| `testdata/uchardet/ja/iso-2022-jp.txt` | ISO-8859-1 | ISO-2022-JP |
| `testdata/wpt/legacy-mb-japanese/iso-2022-jp/iso2022jp_chars.html` | ISO-8859-1 | ISO-2022-JP |
| `testdata/wpt/legacy-mb-japanese/iso-2022-jp/iso2022jp_chars-csiso2022jp.html` | ISO-8859-1 | ISO-2022-JP |
| `testdata/uchardet/ko/iso-2022-kr.txt` | ISO-8859-1 | Ascii |
| `testdata/wpt/legacy-mb-japanese/iso-2022-jp/iso2022jp_errors.html` | ISO-8859-1 | Ascii |

The last two are 7-bit files this package cannot decode as ISO-2022: the KR file has no decoder, and the `_errors.html` fixture is deliberately malformed ISO-2022-JP, which chardet reports at 0% confidence. Both decode to the same bytes as before, since the ASCII decoder path is `iso88591`. I left the `Should be ISO-2022-JP?` note on the `_errors.html` entry in place.

`go test ./...`, `go vet ./...` and `gofmt -l` are clean, and the reproduction above now exits 0.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


klaernie (approved):
Looks good to me, I think we iterate with every PR closer to the ideal solution.

theoludwig (approved):

